PR #6, opened 2021-06-11
Add checks for installed programs

paniash:


pystardust:
Yes, we need to add the dependency check, but using which is very suboptimal, since it is calling a subshell and then we are testing for empty string, its better to do it with `command -v`

paniash:
Awesome.

PR #8, opened 2021-06-11
Added dmenu support

paniash:
If the user does not provide any argument, a dmenu prompt is shown to search for anime.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -32,7 +32,7 @@ die () {
 search_anime () {
 	# get anime name along with its id
 
-    # If no argument is passed use dmenu prompt
+    # If no argument is passed, use dmenu prompt
     if [ -z "$1" ]; then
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -98,7 +109,7 @@ done
 
 shift $((OPTIND - 1))
 
-[ -z "$*" ] && { help_text ; die "Search query not provided";  }
```
paniash, line 101: This line could be entirely removed but it's your call.

pystardust:
Hey @paniash thats nice, but it would be nice if the whole anime selection is also done in dmenu. This would still require the terminal open for selecting anime etc

paniash:
Yeah, makes sense. I'll try adding that too if possible.

ura43:
This seems very stale and has no issue to track the Feature request. Please reimplement and open another PR if the feature is still desired

Issue #5, opened 2021-06-10
[Feature-Request]Ability for episode count to continue where we left off


palwd8:
Would it be possible for ani-cli to select the last episode we were on and play that automatically?

pystardust:
What if we store the history in a file.
Like each anime name with the last episode the user has watched.
So when a user does ani-cli -H
Then we could show the latest episodes of the currently watching anime

pal-wal:
Oh yeah that'd work as well

PR #14, opened 2021-06-14
feat: episode count to continue where we left off

Dink4n:
#5


pystardust:
Thats really nice having the history feature like that. Would it be better to use a tsv(  tab separation) instead of using json.
Since its only key value pairs, I think it would be simpler to use the anime-id and episode number tab separated, each of this on a new line

Dink4n:
I have currently replaced json with tsv.
I haven't rebased

PR #16, opened 2021-06-17
feat:start playing from where we left off

palwd8:
if the currently selected anime is an old anime then it automatically starts playing from where we left off


pystardust:
refer #14

PR #17, opened 2021-06-18
Reset shell color when exiting via keyboard interrupt

r2dev2:
Previously, when you control c in the script, the shell will start to be displayed in green. This echos the reset color code when there is a keyboard interrupt.


pystardust:
yes that a nice consideration. But echo -e is a bashism, try using printf

PR #23, opened 2021-07-07
Update ani-cli

ViktorMYT:
I think vlc would be better, mpv is lagging while vlc works smoothly with any video file you can imagine, while mpv will complain if you don't have proper codecs installed so I think that vlc is better in this case.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -117,7 +117,7 @@ play_ep () {
 	done<<-!
 	$html
 	!
-	mpv "$url"
+	vlc "$url"
 }
 
 case $selected_anime_url in


pystardust:
thanks personal preference I guess, we could put that as a variable/function instead of hardcoding it maybe

J0hsHH:
--mpv
--vlc

Shinyzenith:
> Better alternative, use `xdg-open` this would let the user specify how to open files.

`xdg-open` could potentially call on a video player that can't render URL's? Luckily MPV does. Personally i'd prefer setting my own video player in an environment variable or as a variable in the script.

pystardust:
the video player can be set by changing the player_fn variable in the start of the script

PR #33, opened 2021-09-19
fix: change to the new Referer header link

### LINKED ISSUE #32: mpv not opening
I can search for shows just fine, but when I try to play an episode it doesn't open. When I try to download an episode the mp4 file data reads as follows:

```
<html>
<head><title>403 Forbidden</title></head>
<body bgcolor="white">
<center><h1>403 Forbidden</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

I already implemented @Almo7aya 's fix to bypass gogoanime's checks from a week ago, but this didn't change anything.

Almo7aya:
Fix #32 

gogoanime.vc changed the link of the Referer header to https://goload.one/, I think they will keep changing it everytime we fix, I guess we scrape it from somewhere and append, since it's a custom header so we can easily get it

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -206,7 +206,7 @@ open_episode () {
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new" && mv "${logfile}.new" "$logfile"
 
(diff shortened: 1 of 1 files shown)


Rafonel:
upvote!

Dink4n:
The variable dpage_url is the referer

ura43:
Seems to be an earlier instance of the url fixing done by more recent PRs.
Closing

PR #35, opened 2021-10-07
fix: download feature

Dink4n:
Fixed the download feature
There is an extra dependency (ffmpeg)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,3 +32,4 @@ This would open/download episodes 1 2 3 4 5 6
 * curl
 * sed
 * mpv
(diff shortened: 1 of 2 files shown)


pystardust:
I think we need to add this to the docs @Dink4n

ganesh314159:
@pystardust when are you going to pull his request?

AlexFreik:
@pystardust  Do you not want to pull this request because of extra dependency?

pystardust:
@Dink4n feel free to push directly. I may not be available due to academics.

PR #41, opened 2021-10-16
Added in the ability to accept URLs from gogoanime as a query, as well as to run non-interactively for scripting.

aflyingpumpkin:
These features allow users to better interact with this utility via other bash scripts.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -105,7 +105,7 @@ get_search_query () {
 	else
 		query=$*
 	fi
-	if [ ! "$(expr "$query" : "https://gogoanime\..*")" = 0 ]; then
+	if printf '%s' "$query" | grep -q "https://gogoanime\.."; then
 		query_is_url=1
 	else
 		query_is_url=0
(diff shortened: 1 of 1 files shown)


pystardust:
Try changing the indentation to match the single tab per fold as per rest of the code to keep the commits clean

ura43:
Conflict is easy enough to resolve

aflyingpumpkin:
What all do I need to do to get this merged?

ura43:
> What all do I need to do to get this merged?

Wait
And not even that long. Will probably merge today

ura43:
Actually, I don't think you'll get this merged.
Accepting URLs bloats the code in an unacceptable way, that feature will not make it in.
Non interactiveness on the other hand is a decent feature, but without URL passing it is meaningless, as your queries get passed to the search engine and only interactively can you select your choice.
I understand that you might want to make a script that downloads a large amount of anime.
But there are dozens of gogo scrapers out there and you can probably even write your own if you truly have a usecase

aflyingpumpkin:
I actually did write my own a while back, my use case is taking an RSS feed as an input and automatically downloading new episodes.

PR #52, opened 2021-10-17
added instructions for windows

purplevarun:
added instructions in readme.md as to how to run the program in windows using cygwin.


pystardust:
windows has a separate branch, push over there

PR #53, opened 2021-10-17
Easy Windows instructions added

avidhy-chan:
Easy Windows instructions added with fewer steps to get started in less time and few commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,3 +44,13 @@ This would open/download episodes 1 2 3 4 5 6
 * curl
 * sed
 * mpv
+
+## Windows install instructions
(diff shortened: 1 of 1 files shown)


pystardust:
Yup the instructions are added to the new branch, push any updates to that
@avidhy-chan I have given you access to push

ura43:
It has been a while since I last used windows, but I believe it would be bad practice to suggest a user to use a package manager when they most likely have all other programms installed through exe's from the internet.
Even if suggesting a package manager were right, I don't see why scoop would be inherently better than lets say chocolatey or even the windows store

ura43:
Fixed in #234

PR #55, opened 2021-10-17
Previous/next episode menu works on keypresses

Lukasz825700516:
Selection of next/previous episodes that shows up when opening an video,
now works by pressing single keys, without requirement of pressing
enter key.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -27,6 +27,7 @@ help_text () {
 	 -D	 delete history
 	 -q	 set video quality (best/worst/360/480/720/..)
 	 --dub  play the dub version if present
+	 -k 	 on keypress navigation (previous/next/replay/quit episode)
 	EOF
 }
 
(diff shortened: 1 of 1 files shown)


pystardust:
This is a POSIX sh script, read with those flags is a bash extension

GMAGD:
hi, i've create a version that works with VLC 💯

pystardust:
this is a cool feature but I feel the defaults are cleaner. Since if you press the key on accident you have a chance to correct  yourself

Lukasz825700516:
Then should this be an option that would turn on with some flag like `-k` (key presses)?

pystardust:
sounds good :D

ura43:
> This is a POSIX sh script, read with those flags is a bash extension

Thats why shellcheck will be a thing soon

ura43 (approved):
Genius idea, I doubt I'll use anything but -k once I merged it.
The conflicts are rather simple to resolve. Leave that to me

PR #61, opened 2021-10-18
Update README.md

aditya6421:
ani-cli-win instead of ani-cli


pystardust:
changes are already made

PR #62, opened 2021-10-18
Added dmenu as an interface option

cimelo:
I added dmenu as a full handler for the script.


pystardust:
Great, but this is clashing with the quality path which was patched before this

PR #91, opened 2021-10-28
New delete history option.

joao-vitor-sr:
I added this option for the user to choose which anime to delete from the history.
Using the `-D` option


pystardust:
this adds a lot of complexity to the code, since the history is just a text file it would be simpler to just delete the single line. But deleting the whole history seems good (the first commit).
```sh
: > $file
```
You don't need a subshell, you could do this instead

PR #92, opened 2021-10-28
Added the possibility to search for another anime in the main loop

Derisis13:
I had this annoyance with your script that I need to relaunch it if I wish to watch another series.
To solve this I created a new option to the main loop: entering "a" will now bring up the search anime again. (I used cyan, because I don't know the reason for the original colour choices. Change it if you feel like so.)
image
I copied the code that runs on startup (query mode) since I don't understand the script that well. There might be better ways to do it, but this appears to work just fine.
It would make sense to make a function for those four lines, but I couldn't come up with an intuitive enough name for them. It shouldn't take more than a minute though.
I have tested my modifications myself, but I think an independent test would be appropriate to ensure I didn't break the script.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -371,11 +371,14 @@ while :; do
 			episode=$((episode))
 			;;
 		a)
-			get_search_query "$*"
+			get_search_query ""
 			search_results=$(search_anime "$query")
 			[ -z "$search_results" ] && die "No search results found"
 			anime_selection "$search_results"
-			episode_selection;;
+			episode_selection
+			grep -q -w "${selection_id}" "$logfile" ||
+				printf "%s\t%d\n" "$selection_id" $((episode+1)) >> "$logfile"
+			;;
 
 		q)
 			break;;
--- a/ani-cli
+++ b/ani-cli
@@ -356,19 +356,23 @@ while :; do
 	case $choice in
 		n)
 			episode=$((episode + 1))
+			open_episode "$selection_id" "$episode"
 			;;
 		p)
 			episode=$((episode - 1))
+			open_episode "$selection_id" "$episode"
 			;;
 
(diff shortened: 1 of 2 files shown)


Derisis13:
Feature isn't complete: 
Issue: If you search for an anime that was already played in this instance of ani-cli the episode number doesn't get overwritten.

Derisis13:
> Feature isn't complete: Issue: If you search for an anime that was already played in this instance of ani-cli the episode number doesn't get overwritten.

The last two commits fixed this one, the feature is ready for merging to master

ani-cli
```diff
@@ -348,26 +354,41 @@ while :; do
...
 			[ "$episode" -eq "$episode" ] 2>/dev/null || die "Invalid number entered"
+			open_episode "$selection_id" "$episode"
 			;;
 
 		r)
 			episode=$((episode))
+			open_episode "$selection_id" "$episode"
+			;;
+		a)
+			get_search_query ""
```
Dink4n, line 384: Clearing the screen before asking for search query would be good ;)

ani-cli
```diff
@@ -348,26 +354,41 @@ while :; do
...
+			search_results=$(search_anime "$query")
+			[ -z "$search_results" ] && die "No search results found"
+			anime_selection "$search_results"
+			episode_selection
+			check_input
+			append_history
+			open_selection
```
Dink4n, line 391: You could add a `continue` statement here instead of adding `open_episode` for every choice.

ani-cli
Derisis13: Fixed in the new commit. I didn't do it initially because it doesn't do that on startup, but I was thinking about it.

ani-cli
Derisis13: Fixed in the new commit. I wish I did it like this for the first time.

PR #96, opened 2021-10-29
Added flag for specifying download directory

HumanMan007:
As seen in some issues some people want to specify the download directory (which would be rather useful if downloads where working...) in this patch I added the ability to specify the download directory via the option -p (standing for path).
I didn't put it as the $optarg for -d as that seemed rather annoying for those that don't need it.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -250,13 +250,14 @@ open_episode () {
 
 		setsid -f $player_fn --http-header-fields="Referer: $embedded_video_url" "$video_url" >/dev/null 2>&1
 	else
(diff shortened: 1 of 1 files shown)


ura43:
Seems good in theory, lets see if it eventually does

ura43:
Conflicts are easy to resolve

ura43:
Just noticed: its really stupid to have another flag for this when that could be an optional argument for -d

muhi03:
Screenshot 2022-01-03 at 00 25 13


This helped for me...

PR #104, opened 2021-11-03
streaming in dubbed format

r4vanan:
listening anime in English is cool


pystardust:
thanks, but it would be better to add a flag instead of having dub as default

PR #105, opened 2021-11-03
Added ability to automatically save episode playback and resume it

JLO64:
NGL, I'm the type of person who has issues watching a full episode in one sitting. I usually pause or stop in the middle because I'm a zoomer (or millennial). This became such an issue for me that I stopped torrenting anime and switched to official streaming sites because they tracked playback progress. No more.

Now with the power of God(Lain) and Anime on my side I've added the ability for this program to remember your playback progress in the same file as your viewing history and to resume it with the '-H' option. This works by having MPV write to an IPC socket its playback progress while a background process checks playback progress every 10 seconds and writes it to the logfile. The point it stops recording your progress is either when you close the MPV player or the program itself. This solution is specific to MPV but in the future i might work on a VLC version of this.

The biggest issue with this method of playback recording is that it requires two dependencies to work, _socat_ and _jq_. Given that VLC and Windows support is something that is being worked towards, I have added flags to disable this feature to mitigate any potential issues. In short, **if MPV is not being used or if the optional dependencies are not installed, this program will work the same way it has in the past**. There is still a warning that is shown when the optional dependencies are not installed, but it does not terminate the program as it does with non-optional dependencies.

Also, I imagine that people have already been using this program for a while and may have built up their history and don't want it overwritten by a new format. As such I haven't fundamentally changed the format of the logfile that has been used (data separated by tabs) and this program can still read files with the old data format. **So your old anime viewing history will still be saved and won't be overwritten.** I have had to rewrite the way data from the logfile is read but this new method of using arrays could be useful for storing additional parameters unique to each episode/anime such as video quality or dub/sub.

Finally, once playback reaches the last 15 seconds of an episode the background process will switch over from writing the current playback location of the current episode to writing the starting location (0 seconds) of the next episode. I noticed that when I didn't have this set, using '-H' resulted in just the last 10 seconds of a show playing and MPV closing immediately which could be frustrating to some users. I am considering extending this time period from 15 seconds to a minute but for now I'll leave it here.

As always, sorry for my sloppy code!

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -24,8 +24,9 @@ help_text () {
 	 -h	 show this help text
 	 -d	 download episode
 	 -H	 continue where you left off
-	 -D	 delete anime from history
+	 -D	 delete history
 	 -q	 set video quality (best/worst/360/480/720/..)
+	 --dub  play the dub version if present
 	EOF
 }
 
@@ -71,7 +72,8 @@ get_embedded_video_link() {
 	ep_no=$2
 
 	# credits to fork: https://github.com/Dink4n/ani-cli for the fix
-	curl -s "https://gogoanime.pe/$anime_id-episode-$ep_no" |
+	# dub prefix takes the value "-dub" when dub is needed else is empty
+	curl -s "https://gogoanime.pe/$anime_id${dub_prefix}-episode-$ep_no" |
 	sed -n -E '
 		/^[[:space:]]*<a href="#" rel="100"/{
 		s/.*data-video="([^"]*)".*/https:\1/p
@@ -278,7 +280,7 @@ dep_ch "$player_fn" "curl" "sed" "grep"
 is_download=0
 quality=best
 scrape=query
-while getopts 'hdHDq:' OPT; do
+while getopts 'hdHDq:-:' OPT; do
 	case $OPT in
 		h)
 			help_text
@@ -297,6 +299,13 @@ while getopts 'hdHDq:' OPT; do
 			;;
 		q)
 			quality=$OPTARG
+			;;
+		-)
+			case $OPTARG in
+				dub)
+					dub_prefix="-dub"
+					;;
+			esac
 			;;
 	esac
 done
--- a/ani-cli
+++ b/ani-cli
@@ -129,7 +129,15 @@ get_links () {
 dep_ch () {
 	for dep; do
 		if ! command -v "$dep" >/dev/null ; then
-			die "Program \"$dep\" not found. Please install it."
+			die "Program \"$dep\" not found. Please install it.\n"
+		fi
+	done
+}
+
+optional_dep_ch () {
+	for optional_dep; do
+		if ! command -v "$optional_dep" >/dev/null ; then
+			printf "Optional program \"$optional_dep\" not found. It is recommended to install it.\n"
 		fi
 	done
 }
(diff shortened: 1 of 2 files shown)


pystardust:
Wow, I didn't know mpv could be controlled this way.
This is a posix shell script and not a bash script. We could either port your changes into posix compatible script or convert the script to a bash script. @Dink4n what's your opinion?

JLO64:
_Frantically searches the web about what a posix shell script is..._

@pystardust @Dink4n Ah $h!t, that was my bad. If I'm gonna be honest I'd prefer to keep it as a posix script and not a bash script. Thankfully my linter isn't throwing that many posix related warnings so hopefully it shouldn't be that hard to port my changes. I'll try to get it finished within a day or so (I have a bunch of free time and I'm down to try and learn something new), but if I have any issues I'll post another comment here.

Also just as a suggestion, I've noticed others making my same mistake so maybe we should mention that this is a posix script somewhere in the README?

Dink4n:
@pystardust Isn't it better to keep the script posix compatible

JLO64:
@pystardust @Dink4n After figuring out how to rebase (and screwing it up) and losing my progress and starting from scratch, the script should be okay now. My linter (Shellcheck) isn't thowing any POSIX related warnings anymore.

JLO64:
On second thought, given how messy my commit history is after I screwed up my rebase, just say whether or not my changes are acceptable to be pushed to main. I'll then close this PR and start a new fork and copy my changes to that one and create a new PR based off of that. That way we can keep our commit history tidy.

pystardust:
> How about some kind of "Contributing Guidelines" section in README

I have added a simple contributing.md, feel free to edit

ani-cli
```diff
@@ -248,7 +249,12 @@ open_episode () {
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new" && mv "${logfile}.new" "$logfile"
 
-		setsid -f $player_fn --http-header-fields="Referer: $embedded_video_url" "$video_url" >/dev/null 2>&1
+		#In order for playback history to be recorded MPV needs to be used
+		if [ $player_fn == "mpv" ]; then
```
pystardust: we are using `http-header-fields` option which makes this script only work with `mpv`. So no need to do this check for now

ani-cli
```diff
@@ -350,6 +363,19 @@ while :; do
 	printf "$c_blue[${c_cyan}%s$c_blue] $c_magenta%s$c_reset\n" "r" "replay current episode"
 	printf "$c_blue[${c_cyan}%s$c_blue] $c_red%s$c_reset\n" "q" "exit"
 	printf "${c_blue}Enter choice:${c_green} "
+
+	#Start a background process that checks for playback progress and writes it
+	if [ $player_fn == "mpv" ]; then
+		while true
+		do
+			sleep 10s
+			POSITION=$(echo '{ "command": ["get_property_string", "time-pos"] }' | socat - /tmp/mpvsocket 2>/dev/null | jq .data | tr '"' ' ' | cut -d'.' -f 1)
```
pystardust, line 398: you could use `jq` with the `-r` flag to remove quotes, no need of extra `tr`

ani-cli
```diff
@@ -371,12 +397,15 @@ while :; do
 			;;
 
 		q)
+			#first kill the background process checking for playback progress then break
+			jobtokill=$(jobs -l | grep "Running" | grep -o "[[:digit:]]\{3,10\}")
```
pystardust, line 433: there is no `jobs` command in posix shell so we would need to have a pidfile for the running bg process. That we read a kill later on

pystardust:
@JLO64 ill open a branch for this called save-resume we can work over there before merging it. Try sending a pr to that branch

ani-cli
JLO64: Damn. My linter probably didn't catch this since it's inside the $()

JLO64:
I'll be busy for a while but I'll send a PR to that branch as soon as I can. One last question, is there any reliable way to check if a script is posix compliant? I've tried running bash with --posix but that seems to have no effect.

ani-cli
JLO64: Should we leave it in for future use? I don't think it makes much of a difference either way.

Issue #106, opened 2021-11-04
MPV not opening after selecting Ep - OpenSuse


santi014:
I have all the dependencies installed but its not opening.
#43 Pretty similar case but for Linux in this case.
![Screenshot_20211104_024740](https://user-images.githubusercontent.com/80226768/140264294-4332b7b9-4c6c-47c6-bb66-7d0f457c7486.png)
New to github sorry for any mistake

Dink4n:
What is your version of `setsid`?
```Console
setsid -V
```

santi014:
Hi its, util-linux 2.36.2

Dink4n:
Add this flag `--log-file=debug.log` just before `--http-header-fields` in the script
https://github.com/pystardust/ani-cli/blob/babcc68a48705b67f593e2775286a95dcf2b072e/ani-cli#L255

Try to use `ani-cli` now. It should create a `debug.log` file
Can you print the contents of the file.

kd00r:
I think I got some valuable information for you. I have the exact same problem as santi014. I added the flag `--log-file=debug.log` but the log was not to be found anywhere. After about 45 minutes of troubleshooting, I landed on changing the line to `setsid $player_fn --http-header-fields="Referer: $embedded_video_url" "$video_url" >/dev/null 2>&1 `

By removing the `-f` it now works properly for me. `-f ` doesn't seem to be a valid flag for `setsid`.

I have version `setsid from util-linux 2.31.1` running `Ubuntu 18.04` if that helps.

santi014:
> Add this flag `--log-file=debug.log` just before `--http-header-fields` in the script
> 
> https://github.com/pystardust/ani-cli/blob/babcc68a48705b67f593e2775286a95dcf2b072e/ani-cli#L255
> 
> Try to use `ani-cli` now. It should create a `debug.log` file Can you print the contents of the file.
This is my debug.log
[debug.log](https://github.com/pystardust/ani-cli/files/7477617/debug.log)





> I think I got some valuable information for you. I have the exact same problem as santi014. I added the flag `--log-file=debug.log` but the log was not to be found anywhere. After about 45 minutes of troubleshooting, I landed on changing the line to `setsid $player_fn --http-header-fields="Referer: $embedded_video_url" "$video_url" >/dev/null 2>&1 `
> 
> By removing the `-f` it now works properly for me. `-f ` doesn't seem to be a valid flag for `setsid`.
> 
[...]

PR #108, opened 2021-11-04
Update ani-cli to include setsid version check

kd00r:
Please edit line 7, the variable setsid_version should contain the version where the -f parameter was added.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -2,6 +2,10 @@
 
 # dependencies: grep sed curl video_player
 # video_player ( needs to be able to play urls )
+
+verchk () { echo "$@" | awk -F. '{ printf("%d%03d%03d%03d\n", $1,$2,$3,$4); }'; }
+setsid_version="2.36.2" #edit this. I do not know when the -f parameter was introduced.
+
 player_fn="mpv"
 
 prog="ani-cli"
@@ -14,7 +18,6 @@ c_blue="\033[1;34m"
 c_magenta="\033[1;35m"
 c_cyan="\033[1;36m"
 c_reset="\033[0m"
-
 
 help_text () {
 	while IFS= read line; do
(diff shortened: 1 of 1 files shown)


JLO64:
What exactly does this do? I'm only asking since it conflicts with my PR.

kd00r:
It checks the installed version of setsid. If the installed version is below the version specified on line 7, `setsid_version="2.36.2"` in this case, the program never adds the `-f` version to the mpv play command at line 260. If it is above, it adds the `-f`.

There is a bit of trickery going on at line 258, where `sed s/[^0-9.]//g` only returns numbers and dots from `setsid- V` to the added `verchk` bash function on the same line. The only remaining trickery is on line 6, where the `verchk` function is defined. That function, together with the `-ge` (greater than or equal compare) on line 258, compares the version string we got from `setsid -V` vs what is in the `$setsid_version` variable.

The reasoning behind me doing this is because if this would have been in place when I first installed ani-cli, it would have worked out of the box for Ubuntu 18.04 with setsid version `2.31.1`.

Please see issue: https://github.com/pystardust/ani-cli/issues/106 for the full story, it turned out we had two different issues, non-installed decoders vs too early version of setsid.

I hope this makes a bit more sense now!

JLO64:
Thanks for the insightful write up! The things we do to watch anime...

kd00r:
Yeah, indeed. I did not know beforehand that Ubuntu 18.04 (which I unfortunately is forced to use on that particular machine) ships with this old version of setsid. I understand it's purpose in the script, but in my case with the old version did not even spew out the log file due to mpv never starting. The error was just a few characters to the left on the same line, but it took me a good 45min to find it.

pystardust:
setsid is only used to put `mpv` in the background as a forked process.

We could remove `setsid` by just doing 
```sh
mpv "$url" >/dev/null  2>&1 &
```
But a caveat would be closing the terminal will close mpv

ura43:
A move from setsid to nohup is planned to better support mac users and to be closer to posix compliance.
That transition will make this PR obsolete

PR #113, opened 2021-11-06
fix: added progress bar while using -d 

r4vanan:


r4vanan:
now its using python module to give process bar

ura43:
The additional dependency might make this unappealing. At least it should be optional in some way and explicitly requested if you know what I mean

ura43:
Could you add error handling for when the progress bar python module isn't installed / make it fall back to not using the feature if no progress bar is present?

ura43 (changes requested):
- `L286` Check via dep_ch if ffpb is installed
- If it is, execute your code block
- If it isn't, execute the current code block
- Remove the pip install from the makefile

ura43:
I'll soon™ work on a change to the README that enables seperate listing of optional dependencies with the feature they enable and add yours there

ura43:
Fixed with #165 
Well, the progress indicator. It's not a bar, but gives you all the stats you want to know during download

Issue #117, opened 2021-11-06
External runscripts for starting mpv


Lukasz825700516:
I don't often switch episodes while still watching current one, so separate terminal window bugs me a little.

So I would like to add `-r runscript` option that would allow starting mpv through user defined runscript. In my case i would use this option with devour, or other "hiding current window until program exits" program.

This however would assume that some of runscripts would not fork to background, as it currently is.

Derisis13:
Would changing the variable `player-fn` in line 5 solve your problem? If so, it'd be super simple to add an extra argument.

Lukasz825700516:
Like `player-fn="run_script_and_mpv.sh"`? I think it would be a bit inconvenient as then I would need to have these custom made scripts in some known location, and that could collide with future features that could possibly set player-fn to something else + that would make eventual porting into windows more harder.

Derisis13:
I was thinking taking a -r argument that contains the video player too, but what you did in #120 is actually better.

ura43:
How is that PR not merged yet? 
Remind me in like a week if I haven't merged it by then

ura43:
But yes, having a player option is a good idea

ura43:
> But yes, having a player option is a good idea

It's not if you think it through. The code should be easy enough to be easily modifyable by anyone who has the competence to write his external runscript

PR #120, opened 2021-11-07
Added option to specify video player launcher

Lukasz825700516:
Now setsid -f can be replaed by user with something else

This is my proposition of #117 (tldr; allow user modify current terminal window on video launch)

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -3,6 +3,8 @@
 # dependencies: grep sed curl video_player
 # video_player ( needs to be able to play urls )
 player_fn="mpv"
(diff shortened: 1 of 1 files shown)


ura43:
Looks good to me

ura43:
This wasn't the case when you were writing this PR, but ani-cli supports vlc now.
I don't think a custom script should be supported.
By explicitly telling the user what we support, mpv and vlc currently, we avoid solving unnescessary issues.
And if a user is experienced enough to write his own player script, he can always modify the source

PR #121, opened 2021-11-07
Change copy to symlink in Makefile

sergioguaka:
I changed `cp` for `ln -sr`, which creates a relative symbolic link instead of a copy. 

With `cp` it's necessary to run `make install` after pulling and it's possible to have different versions of the script in the repo folder and `usr/local/bin/`.

The symbolic links only needs to be done once and it will make unnecessary to execute `make install` after every update, since any changes to the repo file will affect the one in `usr/local/bin` as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ PREFIX := /usr/local
 all: install
 
 install:
-	cp ani-cli $(DESTDIR)$(PREFIX)/bin/ani-cli
+	ln -sr ani-cli $(DESTDIR)$(PREFIX)/bin/ani-cli
 	chmod 0755 $(DESTDIR)$(PREFIX)/bin/ani-cli
 
 uninstall:


3ricsonn:
I think this should be optional as some users may not want to keep the cloned directory
Something like:
```makefile
install-link:
           ln -sr ani-cli $(DESTDIR)$(PREFIX)/bin/ani-cli
           chmod 0755 $(DESTDIR)$(PREFIX)/bin/ani-cli
```

ura43:
I propose you make that a hardlink. Solves the issue of deduplication but doesn't break when the user removes the cloned directory

ura43:
Since we have update checking now, ani-cli would break without the cloned repo anyways.
Symlinking would be a nice way to make sure that ani-cli doesn't execute at all when the repo isn't present

Issue #124, opened 2021-11-08
No search results found


VSuki3581:
It was fine a couple days ago but for some reason when i tried to watch naruto today because i was feeling nostalgic it told me no search results found. I already reinstalled ani-cli and restarted my system. No dice, I'm really confused because everything seems like it should be working.
Thank you in advance.

duncanjake1:
Also having this issue. I pulled the latest commits from this repo, but still not getting any search results when I run.

duncanjake1:
@VSuki3581 can you open up the ani-cli file on your machine and see if the following line is included? Should be on line 9.
```
base_url="https://gogoanime.cm"
```
It looks like the gogoanime URL they were using in previous versions no longer works, so they had to change over to this one a couple days ago. If that line isn't present, try running 
```
git pull
```
followed by
```
sudo make
```

Anky209e:
I was also facing this just pull the latest changes by above command. It will work fine.

VSuki3581:
Thanks! It works now :D

teymurov27:
thanks it worked xD

ArekkusuMagni:
I am getting an error  when I issue the git pull command. Any help would be appreciated. 

$ git pull
fatal: not a git repository (or any of the parent directories): .git

duncanjake1:
@ArekkusuMagni Are you issuing the command while inside the ani-cli directory?

ArekkusuMagni:
@duncanjake1 Thank you so much, as soon as you posted that question it clicked in my mind that I need to "cd ani-cli" first. 

instructions for future users: 

cd ani-cli
git pull
sudo make

King-Crimson76:
@ArekkusuMagni when i type in "sudo make" it says:
sudo: make: command not found

duncanjake1:
@King-Crimson76 Seems like 'make' isn't installed by default on your disto of linux (or it got removed somehow). Look up how to install it on your distro and try again.

King-Crimson76:
@duncanjake1 Installed it, typed in cd ani-cli and then sudo make since i already did git pull, this came up:

cp ani-cli /usr/local/bin/ani-cli
chmod 0755 /usr/local/bin/ani-cli

duncanjake1:
@King-Crimson76 That means it worked! Try running the script.

King-Crimson76:
@duncanjake1 Yep! It did! Thank you so much!

SaleeSlx:
[...]

PR #126, opened 2021-11-09
Inform user if update is available. Includes instuctions on how to update. Added Git to dependencies.

duncanjake1:
Updated ani-cli to check if the local repo is up to date with the remote repo on script start up. Included a message to be echo'd when an update is available or the local branch has become diverged.

Here's a nifty SO that explains the git syntax - https://stackoverflow.com/a/3278427/16004619

Added git to required dependencies in README.md and dep_ch method.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,5 +1,5 @@
-# Used to track where .git directory is. Variable must be on line 2. Makefile writes to this line.
-ORIGINAL_DIR=/home/duncan/Apps/ani-cli2-fork
+#Used to track where .git directory is. Variable must be on line 2. Makefile writes to this line.
+ORIGINAL_DIR=0
 #!/bin/sh
 
 # dependencies: grep sed curl video_player
@@ -24,7 +24,6 @@ help_text () {
 		printf "%s\n" "$line"
 	done <<-EOF
 	USAGE: $prog <query>
-	-uuu go away
 	 -h	 show this help text
 	 -d	 download episode
 	 -H	 continue where you left off
--- a/ani-cli
+++ b/ani-cli
@@ -311,9 +311,10 @@ open_episode () {
 # to clear the colors when exited using SIGINT
 trap "printf '$c_reset'" INT HUP
 
-check_for_update
 
 dep_ch "$player_fn" "curl" "sed" "grep" "git"
+
+check_for_update
 
 # option parsing
 is_download=0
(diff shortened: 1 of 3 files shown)


duncanjake1:
Created this ticket in reaction to issue #124, Having this feature might help to clear up some confusion in the future.

ppcpwn (approved):

Beyley (approved):
lgtm

ani-cli
```diff
@@ -127,6 +127,19 @@ get_links () {
 	printf '%s' "$video_url" | sed -n -E "s/(.*)\.m3u8/\1.$video_quality.m3u8/p"
 }
 
+check_for_update () {
+	git fetch 
+	UPSTREAM=${1:-'@{u}'}
+	LOCAL=$(git rev-parse @)
+	REMOTE=$(git rev-parse "$UPSTREAM")
+
+	UPDATE_MESSAGE="UPDATE AVAILABLE\nPlease run the following commands in your ani-cli directory to update\n----------\ngit pull\nsudo make\n----------\n"
```
Dink4n, line 136: Is there a need for UPDATE_MESSAGE since we only use it once?

ani-cli
```diff
@@ -127,6 +127,19 @@ get_links () {
...
+
+	if [ $LOCAL != $REMOTE ]; then
+    echo $UPDATE_MESSAGE
```
Dink4n, line 139: Can you fix the indentation here? Also, instead of `echo`, use `printf`.

ani-cli
Dink4n, line 136: Is there a need for `UPDATE_MESSAGE` since we only use it once?

duncanjake1: I removed the variable and made the text inline with printf

ani-cli
duncanjake1: done and done

duncanjake1:
I noticed that when the ani-cli command was ran outside of the ani-cli directory, the git fetch command would also run in that directory. This would end up running git fetch on a git directory that you didn't intend to, or printing out error text. 

To solve this, I added the ability to write the present directory into a variable in the ani-cli file at the moment that 'sudo make' is ran. git fetch will then be ran on that specific directory, no matter where you are when you run it.

RafaelKuhn (changes requested):
Hey, my man, you shouldn't be PRing this with a path to a plnace inside your computer
line 2: ```ORIGINAL_DIR=/home/duncan/Apps/ani-cli2-fork```
maybe consider inserting that path there dynamically... 
for an example when the user runs make install...

duncanjake1:
@RafaelKuhn That's actually the exact functionality that's happening now! In the last commit, I removed the hard coded path and entered a placeholder value that will be overwritten when the user runs the make file.

PR #137, opened 2021-11-20
Replace setsid with nohup for mac compatibility

Cartmanishere:
`setsid` is the only reason this doesn't work on Mac. 

If we replace it with `nohup`, it works perfectly on MacOS.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -256,7 +256,7 @@ open_episode () {
 				if echo "$status_code" | grep -vE "^2.*"; then
 					printf "${c_red}\nCannot reach servers!"
 				else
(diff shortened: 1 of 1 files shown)


ghyatzo:
another option would be `($player_fn --http-header-fields="Referer: $embedded_video_url" "$video_url" > /dev/null 2>&1 &)`

fgyamauti:
Probably stupid. Still, at least for macOS 10.13.6, I had to manually create an empty file /Users/username/.cache/ani-hsts .

ura43:
I welcome the replacement of setsid through nohup.
Please resolve the conflicts with current master and I'll merge

ura43 (changes requested):
Please resolve the current conflicts

Issue #139, opened 2021-11-22
Links seem to have been outdated or gogoanime has blocked ani-cli?


jnishwanth:
```console
Getting data for episode 111
Downloading episode 111 ...
https://loadfast1.com/www10/c6e4aca092c75849d378f418687a5c58/ep.111.1613285431.720..m3u8
ffmpeg version n4.4.1 Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 11.1.0 (GCC)
  configuration: --prefix=/usr --disable-debug --disable-static --disable-stripping --enable-amf --enable-avisynth --enable-cuda-llvm --enable-lto --enable-fontconfig --enable-gmp --enable-gnutls --enable-gpl --enable-ladspa --enable-libaom --enable-libass --enable-libbluray --enable-libdav1d --enable-libdrm --enable-libfreetype --enable-libfribidi --enable-libgsm --enable-libiec61883 --enable-libjack --enable-libmfx --enable-libmodplug --enable-libmp3lame --enable-libopencore_amrnb --enable-libopencore_amrwb --enable-libopenjpeg --enable-libopus --enable-libpulse --enable-librav1e --enable-librsvg --enable-libsoxr --enable-libspeex --enable-libsrt --enable-libssh --enable-libsvtav1 --enable-libtheora --enable-libv4l2 --enable-libvidstab --enable-libvmaf --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxcb --enable-libxml2 --enable-libxvid --enable-libzimg --enable-nvdec --enable-nvenc --enable-shared --enable-version3
  libavutil      56. 70.100 / 56. 70.100
  libavcodec     58.134.100 / 58.134.100
  libavformat    58. 76.100 / 58. 76.100
  libavdevice    58. 13.100 / 58. 13.100
  libavfilter     7.110.100 /  7.110.100
  libswscale      5.  9.100 /  5.  9.100
  libswresample   3.  9.100 /  3.  9.100
  libpostproc    55.  9.100 / 55.  9.100
[https @ 0x55b0df4fae40] No trailing CRLF found in HTTP header. Adding it.
[https @ 0x55b0df4fae40] HTTP error 404 Not Found
https://loadfast1.com/www10/c6e4aca092c75849d378f418687a5c58/ep.111.1613285431.720..m3u8: Server returned 404 Not Found
Download failed episode: 111
```
[...]

PR #141, opened 2021-11-23
History in loop

Derisis13:
Just like the `a` option in the "menu" can run a query from gogoanime, the new `h` option lists what is in your history file.
Also includes the fix for #139 as suggested by @Dink4n 
I also touched the colour of the `a` option to make it an uniform yellow-magenta theme (it didn't really come trough when I wrote that part originally)

I'm also wondering whether the double semicolons do anything after the continue statements. If they don't should they be removed for decreasing file size and making the code simpler?

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -6,7 +6,7 @@ player_fn="mpv"
 
 prog="ani-cli"
 logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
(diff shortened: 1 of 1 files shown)


Dink4n:
> I'm also wondering whether the double semicolons do anything after the continue statements. If they don't should they be removed for decreasing file size and making the code simpler?

The double semicolons are necessary. Removing them would cause a syntax error.

PR #179, opened 2021-12-08
[hardcoded] Trying to added --wid=WINDOWID

uunnxx:
Hello and thanks for this beatuful cli program!

I don't know much about shell scripting, and I just hardcoded here (I know it'll conflict if someone choice different video player, and hard to change)

So, if you don't mind, would you be able to add some variables like`player_params` at the top of the file, and we would just add  `player_params=--wid=$WINDOWID` to this variable if we need.

We would just leave it empty by default for non-mpv users.


ura43:
Seems like this should go into your mpv config file. Closing, but please open an issue if this is important to you

PR #172, opened 2021-12-05
Adds vlc support

Gotoro:
Uses a somewhat obscure `--adaptive-use-access` tag, which allows VLC to pass the Referer header to all chunks of the video. It doesn't work without it.

Implements a switch statement to possibly add support for some other media players in the future more easily

PR #190, opened 2021-12-12
Allow download without install media player

truboxl:
Skips checking for mpv if user is only downloading

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -132,6 +132,7 @@ get_links () {
 
 dep_ch () {
 	for dep; do
+		[ "$dep" = "$player_fn" ] && [ "$is_download" -eq 1 ] && continue
 		if ! command -v "$dep" >/dev/null ; then
 			die "Program \"$dep\" not found. Please install it."
 		fi
--- a/ani-cli
+++ b/ani-cli
@@ -345,8 +345,9 @@ while getopts 'hdHDq:-:v' OPT; do
 	esac
 done
 
-dep_ch "$player_fn" "curl" "sed" "grep"
 shift $((OPTIND - 1))
+
+dep_ch "$player_fn" "curl" "sed" "grep"
 
 ########
 # main #


ura43:
Seems reasonable to me

ura43 (approved):

ura43:
Just to clarify, does this already implement download support without a media player?

truboxl:
Let me rebase this real quick...
Seems like somebody moved dep_ch already although not to the exact line I wanted...
https://github.com/pystardust/ani-cli/pull/172/files

I will change this PR to only include support download without media player

ura43 (changes requested):
Patch doesn't allow download without mpv installed in the current state.
Maybe depchange isn't late enough yet?

truboxl:
Hmm... weird its working fine for me.
Ok moving dep_ch further back.

ura43:
Oh, I think I was too stupid to checkout your patch-1 branch... and used your master instead.
Feature works

ura43 (changes requested):
On line 321, during the option passing, add a line `player_fn=""`
This makes the check on line 135 redundant

After that you could migrate the 3 `$is_download -eq x` tests to `-z $player_fn` tests
See [here](https://fedingo.com/how-to-check-if-variable-is-empty-or-not-in-shell-script/) if you don't know the -z test.
(tldr: `$x -eq 0` -> `-z $y`, `$x -eq 1` -> `! -z $y`)
Then remove the $is_download variable assignment on 313 and 321.

After that: congrats, you just refactored the -d handling

ura43 (approved):

ura43:
Sorry, that refactor stuff is a bit more ambitious than it initially seemed. I'll merge your PR now

truboxl:
Ah nvm, actually I thought of adding a fallback to download mode as well if mpv is detected not installed if really want to refactor the whole download use case...

PR #197, opened 2021-12-15
Auto-resolve $base_url

MitchMarq42:
This way we can avoid having to hardcode a new base_url every time the domain moves, since they keep the old domains as redirects.

Invokes `curl` once, but may take a bit longer to resolve since it's, well, _resolving_. I used long options here for clarity, but the short ones appear to work as well.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -41,6 +41,10 @@ err () {
 	printf "$c_red%s$c_reset\n" "$*" >&2
 }
 
+resolve_baseurl () {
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -340,6 +344,8 @@ shift $((OPTIND - 1))
 # main #
 ########
 
+base_url=$(resolve_baseurl $base_url)
```
Dink4n, line 347: You don't need to pass `$base_url` to `resolve_baseurl` because it doesn't use it.

PR #209, opened 2021-12-23
Fix base_url and Fix for video not able to play or download

RaynardGerraldo:
Trace gogoanime's base url "https://gogoanime.cm" redirects and getting the final url from redirect as base_url value

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -6,7 +6,7 @@ player_fn="mpv"
 
 prog="ani-cli"
 logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
-base_url="https://www1.gogoanime.cm"
+base_url=$(curl -s -L -o /dev/null -w "%{url_effective}\n" https://gogoanime.cm)
 
 c_red="\033[1;31m"
 c_green="\033[1;32m"
@@ -40,10 +40,6 @@ die () {
 
 err () {
 	printf "$c_red%s$c_reset\n" "$*" >&2
-}
-
-resolve_baseurl () {
-	curl --silent --location --head --output /dev/null --write-out '%{url_effective}' -- $base_url
 }
 
 search_anime () {
@@ -355,8 +351,6 @@ shift $((OPTIND - 1))
 ########
 # main #
 ########
-
-base_url=$(resolve_baseurl $base_url)
 
 case $scrape in
 	query)
--- a/ani-cli
+++ b/ani-cli
@@ -262,7 +262,7 @@ open_episode () {
 	episode_id=$(echo "$embedded_video_url" | grep -oE "id.+?(=&)")
 	video_url="https://gogoplay1.com/download?${episode_id}"
 	
-	play_link=$(lynx -dump -hiddenlinks=listonly $video_url | grep -Eo https:\/\/vidstreamingcdn.com\/.*\/.*\/.*p.* | tail -n 1)
+	play_link=$(wget -qO- $video_url | grep -Eo "(http|https):\/\/vidstreamingcdn.com.*expiry=[0-9]*"| sort -u | sed 's/amp;//' | tail -n 1)
 
 	if [ $is_download -eq 0 ]; then
 		# write anime and episode number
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -6,7 +6,7 @@ player_fn="mpv"
 
 prog="ani-cli"
 logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
-base_url="https://www1.gogoanime.cm"
+base_url=$(curl -s -L -o /dev/null -w "%{url_effective}\n" https://gogoanime.cm)
```
Kungger-git, line 9: This honestly resolves the problem for everyone that gets no results when trying to search for something

shulkk:
Thanks.

shulkk (approved):

RaynardGerraldo:
> Thanks.

np, thanks for the approval🙂

ani-cli
Nikson230: Thanks

solarft (approved):

ScaroDj:
Hello, no longer a fix, I had "ww10.gogoanimehub.tv" and now it is "gogoanimeapp.com", they have always and they will continue to bounce the domain names around. It's the way they evade copyright holders. 

Could there be a way to resolve via DNS to the new URL or even tell the user to resolve it using a web browser and store them in a config file?

RaynardGerraldo:
> Hello, no longer a fix, I had "ww10.gogoanimehub.tv" and now it is "gogoanimeapp.com", they have always and they will continue to bounce the domain names around. It's the way they evade copyright holders.
> 
> Could there be a way to resolve via DNS to the new URL or even tell the user to resolve it using a web browser and store them in a config file?

Did you use (https://gogoanime.cm) for the starting url like in the fix? I just reran the script and it still works and leads me to www2.gogoanime.cm,both from the browser and the script

RaynardGerraldo:
5a0261126682295387977509066f8a6f0ead30f3 Temporary fix for video not able to play or download

OnionHat:
If this PR goes through `lynx` should be added as one of the dependencies in README.

ura43 (approved):
The proposed changes look good to me

ura43 (changes requested):
Please get rid of the lynx requirement and also catch errors for vlc

RaynardGerraldo:
> Please get rid of the lynx requirement and also catch errors for vlc

on it

ura43:
> > Please get rid of the lynx requirement and also catch errors for vlc
> 
> on it

Thanks for tackling this issue. In its current form I still have the same issue. Tried to watch flcl progressive ep 1 on debian unstable. In #231 another user appears to reproduce what I see. 
It definitely says "Video Played", even though it shouldn't

RaynardGerraldo:
> flcl progres

thats good to hear, i removed lynx dependency, vlc error message might take a while to figure out, since it opens vlc even if there is no video to play, havent found a way to output the message before it runs

ura43:
Thats good news. As for vlc, don't bother bending over backwards to make the error display without vlc launching. I'm sure users will close vlc on their own if nothing plays and as long as the error comes after that its good enough.
Wget is already a huge improvement over lynx. If it is not too troublesome, could you adapt it to curl so we have zero new dependencies?
wget and curl are two sides of the same coin iirc so shouldn't be too difficult

ura43 (changes requested):
Great change, but pls investigate if curl instead of wget is practical to avoid new dependency

ura43:
As explained in [this](https://unix.stackexchange.com/questions/593023/equivalent-curl-command-from-wget-command) stackoverflow post, you can just replace `wget -q0-` with `curl -s`

ura43:
Merging and adding the wget to curl substitution right after

RaynardGerraldo:
> Please get rid of the lynx requirement and also catch errors for vlc

ok ill try

> As explained in [this](https://unix.stackexchange.com/questions/593023/equivalent-curl-command-from-wget-command) stackoverflow post, you can just replace `wget -q0-` with `curl -s`



> Thats good news. As for vlc, don't bother bending over backwards to make the error display without vlc launching. I'm sure users will close vlc on their own if nothing plays and as long as the error comes after that its good enough. Copy pasting the stuff you added in the mpv case to the vlc case should do the trick. Wget is already a huge improvement over lynx. If it is not too troublesome, could you adapt it to curl so we have zero new dependencies? wget and curl are two sides of the same coin iirc so shouldn't be too difficult.



> Thats good news. As for vlc, don't bother bending over backwards to make the error display without vlc launching. I'm sure users will close vlc on their own if nothing plays and as long as the error comes after that its good enough. Copy pasting the stuff you added in the mpv case to the vlc case should do the trick. Wget is already a huge improvement over lynx. If it is not too troublesome, could you adapt it to curl so we have zero new dependencies? wget and curl are two sides of the same coin iirc so shouldn't be too difficult.

ok , ill try and do that for vlc. I didnt realize that wget is still an extra dependency, should i open a new pr after this?

Issue #202, opened 2021-12-18
Some episodes wont start playing in mpv


chbo70:
HI, so my main issue is about Black Clover (TV) episodes. After selecting the episode the Player wont start, however i tried it with different animes and they work just fine.

i tried replacing the select video quality print function in issue #192 

but it seems its just this anime...

maybe someone can help me or tell me if they have the same issue
thanks

Duoslow:
i have same issue at "senpai ga uzai kouhai no hanashi"

GetDeleted20454:
same problem

GetDeleted20454:
I tried reinstalling ani-cli and reinstalling the 
 dependincies still doesnt work

alt-art:
I am getting this error when I try to watch One Piece, I tried to do a debug, but as I know little about shell script, it didn't work.

When I tried to download, I realize what is the error:
![image](https://user-images.githubusercontent.com/43281034/146696214-c8825504-6832-4bb2-b6a3-f9f38870e3a4.png)

When I remove the dot in the final of URL, he can get the right m3u file.

`https://www16.anicdn.stream/astream/c751bab1939a2b83020565e1ac242896/one-piece-episode-1004.1639880060..m3u8`
to
`https://www16.anicdn.stream/astream/c751bab1939a2b83020565e1ac242896/one-piece-episode-1004.1639880060.m3u8`

I don't know much about shell scripting, but the error seems to be in `get_links` line 129.

alt-art:
I think he couldn't find the quality of the video.

alt-art:
I will try to make a condition to this at least.

alt-art:
I changed the line 129 with this:
```sh
    # Replace the video with highest quality video
    if [ -z "$video_quality" ];
  	 then printf '%s' "$video_url" | sed -n -E "s/(.*)\.m3u8/\1.m3u8/p";
         else printf '%s' "$video_url" | sed -n -E "s/(.*)\.m3u8/\1.$video_quality.m3u8/p";
    fi
```

Now Black Cover and One Piece works, but some animes may not work yet.

chbo70:
i did replace it however black clover still doesnt work

chbo70:
well, if i am precise, the first episode works but without sound, all the episodes after that still don't work

chbo70:
Here is my 
[debug.log](https://github.com/pystardust/ani-cli/files/7744527/debug.log)

alt-art:
[...]

Issue #135, opened 2021-11-18
Wrong Error Message when connection refused


SaleeSlx:
It worked perfectly when I last used ani-cli about a week ago, but now whatever I search in ani-cli the the output is always "No search results found"

SaleeSlx:
Solved**
In terminal do:

cd ani-cli 
git pull
sudo make

CubeyAlchemist:
I had the same problem, however the version of the repo wasn't the problem for me.

After checking for updates and a few other bits; I tried to access the website normally, and found that it was being blocked by the WiFi provider. Unfortunately this causes the same 'error' message to appear in terminal as a genuine 'no results found'.

If the above fix by SaleeSix doesn't work, try loading gogoanime on your browser. If it has a 'PR_CONNECT_RESET_ERROR', then that means no anime for you my friend.

zergreen:
same problem like you

ura43:
I encountered this due to a rate limit after testing >10 different anime in a short timeframe

ura43:
Error handling should be a lot better now. When connection is refused, the curl error gets printed

PR #232, opened 2021-12-30
Fix regex bug for gogoplay1 link

RaynardGerraldo:
Just tested the script a bunch of times and found a regex bug, updated error messages for both vlc and mpv using status code

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -261,7 +261,7 @@ open_episode () {
 	episode_id=$(echo "$embedded_video_url" | grep -oE "id.+?&")
 	video_url="https://gogoplay1.com/download?${episode_id}"
 	
-	play_link=$(curl -s $video_url | grep -Eo "(http|https):\/\/vidstreamingcdn.com.*expiry=[0-9]*"| sort -u | sed 's/amp;//' | tail -n 1)
+	play_link=$(curl -s $video_url | grep -Eo "(http|https):\/\/.*com\/cdn.*expiry=[0-9]*"| sort -u | sed 's/amp;//' | tail -n 1)
 		
 	status_code=$(curl -s -I $play_link | head -n 1|cut -d ' ' -f2)
 
--- a/ani-cli
+++ b/ani-cli
@@ -259,10 +259,12 @@ open_episode () {
 
 	embedded_video_url=$(get_embedded_video_link "$anime_id" "$episode")
 	
-	episode_id=$(echo "$embedded_video_url" | grep -oE "id.+?(=&)")
+	episode_id=$(echo "$embedded_video_url" | grep -oE "id.+?&")
 	video_url="https://gogoplay1.com/download?${episode_id}"
 	
-	play_link=$(wget -qO- $video_url | grep -Eo "(http|https):\/\/vidstreamingcdn.com.*expiry=[0-9]*"| sort -u | sed 's/amp;//' | tail -n 1)
+	play_link=$(curl -s $video_url | grep -Eo "(http|https):\/\/vidstreamingcdn.com.*expiry=[0-9]*"| sort -u | sed 's/amp;//' | tail -n 1)
+		
+	status_code=$(curl -s -I $play_link | head -n 1|cut -d ' ' -f2)
 
 	if [ $is_download -eq 0 ]; then
 		# write anime and episode number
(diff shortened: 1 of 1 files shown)


ura43 (changes requested):
This definitely enhances error handling.
But I don't believe your fix on line 261 to achieve anything, as it just narrows what passes the regex, maybe you can explain

ura43:
Can't confirm fully functional. Still can't watch flcl progressive, but getting the error message from curl along the lines of "url not found" is a welcome improvement.
Option 1) You rename this PR to exclude "Fix regex bug for gogoplay1 link" and we merge just the error handling improvements
Option 2) You deepdive into the regex

RaynardGerraldo:
> This definitely enhances error handling. But I don't believe your fix on line 261 to achieve anything, as it just narrows what passes the regex, maybe you can explain

ok so yesterday, i was testing the script with one piece episodes, and i noticed that curl was returning no url specified, tracked down the problem and it turned out to be the gogoplay link, so the previous regex "id.+?(=&)" matches the string id=ABCDEFGH=& , this was i think because i faced a link with an ID link this (a "=" sign before "&"), but the one piece link had this id=MTc2NjEx& which did not match with the regex, so i removed the "=" sign from regex "id.+?&" and now it matches both

RaynardGerraldo:
> flcl progressive
 
looks like we got a new problem, turns out video links are not only from vidstreamingcdn, for animes like flcl progressive, it uses lb.loadfast1

RaynardGerraldo:
> Can't confirm fully functional. Still can't watch flcl progressive, but getting the error message from curl along the lines of "url not found" is a welcome improvement. Option 1) You rename this PR to exclude "Fix regex bug for gogoplay1 link" and we merge just the error handling improvements Option 2) You deepdive into the regex

@ura43  just fixed the bug in my end, tested it with both dr stone and flcl progressive, but i might need more direct video links to confirm that my regex match is working,

ura43:
> > This definitely enhances error handling. But I don't believe your fix on line 261 to achieve anything, as it just narrows what passes the regex, maybe you can explain
> 
> ok so yesterday, i was testing the script with one piece episodes, and i noticed that curl was returning no url specified, tracked down the problem and it turned out to be the gogoplay link, so the previous regex "id.+?(=&)" matches the string id=ABCDEFGH=& , this was i think because i faced a link with this ID link (a "=" sign before "&"), but the one piece link had this id=MTc2NjEx& which did not match with the regex, so i removed the "=" sign from regex "id.+?&" and now it matches both

Thanks for the clarification, makes sense in that case

ura43:
Can also confirm working with flcl progressive

ura43:
Confirmed working for 5 other anime, I think its safe to assume you fixed the issue.
But I'll go through the reported not working anime in #202 to further verify

ura43 (approved):

ura43:
Confirmed working with all reported problematic anime. Will merge soon

RaynardGerraldo:
> Confirmed working with all reported problematic anime. Will merge soon


good to know, btw just noticed that the video quality function is not working for this code's current state, i set it to always play highest quality as default, if its necessary ill try to update the get_video_quality function to work with the new updates,

ura43:
So launching `ani-cli -q worst` ignores prefference and just plays the highest quality?

ura43:
I think we should merge and open another issue to rework the -q option, as I believe far fewer users will worry about too high bandwith use over not being able to watch

RaynardGerraldo:
> So launching `ani-cli -q worst` ignores prefference and just plays the highest quality?

sorry to say but it doesnt even play the video, ill look into it,  i also see that i can cut down some unnecessary codes while fixing this because its related

ura43:
Is there a rate limit we may have run into?

ura43:
Because after testing the ~10 anime earlier it also failed to play for me, but now it plays again.

RaynardGerraldo:
> Is there a rate limit we may have run into?

Potentially yes, as mentioned in 5a0261126682295387977509066f8a6f0ead30f3 theres an "Error 1101" which i think is a rate limit,thats also why i added status code to detect the error before playing the video

ura43:
Seems like #135 is a good ticket to track this follow up issue

PR #248, opened 2022-01-01
fix: only update history when video could be played

Malertach:
The history file (logfile) should not be updated, when the "Cannot reach servers" issue is reported. Only update the history file when the video is playing.


ura43 (approved):
Nice catch

Dink4n (approved):
Noice

PR #249, opened 2022-01-01
fix: url checks checking not an url

Malertach:
curl used the function names "get_links" and "get_video_quality" as url. Seen in these warnings from my proxy setup:
```
[proxychains] Strict chain  ...  <my_proxy>  ...  get_links:80 <--socket error or timeout!
[proxychains] Strict chain  ...  <my_proxy>  ...  gogoplay1.com:443  ...  OK
[proxychains] Strict chain  ...  <my_proxy>  ...  get_video_quality:80 <--socket error or timeout!
```

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -117,7 +117,7 @@ get_links () {
 get_video_quality() {
 	
 	get_links
-	video_quality=$(curl -s get_links "$video_url" | grep -oE "(http|https):\/\/.*com\/cdn.*expiry=[0-9]*"| sort -V | sed 's/amp;//')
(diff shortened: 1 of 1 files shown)


ura43:
Doesn't really make sense to me, can you elaborate?

Malertach:
> Doesn't really make sense to me, can you elaborate?

`curl` interprets `get_links` and `"$video_url"` as two different urls.
https://github.com/pystardust/ani-cli/blob/2b40f2dd4abbfd80f2c8c2366757a4a24999918a/ani-cli#L130
But only `"$video_url"` gets expanded to an actual url.
`get_links` stays as string and isn't a useful web address.

 Similar issue here:
https://github.com/pystardust/ani-cli/blob/2b40f2dd4abbfd80f2c8c2366757a4a24999918a/ani-cli#L319

ura43:
Oh, yeah thats pretty bad

PR #252, opened 2022-01-01
Fixed update instructions

IvanCodeslol:
The update instructions didn't tell you to remove `/usr/local/bin/ani-cli` before running make which caused errors.
Also not sure why I added the `-rf` flag when its a binary so you can remove that from the commit.


ura43:
Thats a bug with the current make file and not intended behaviour.
It should be fixed in like 2 hours

PR #253, opened 2022-01-01
replace ffmpeg with aria2c

CoolnsX:
replace ffmpeg with aria2c for downloading episodes for following reasons:-
1. resume downloading capabilities of aria2c..
2. parallel downloading (if server supports, not in our case right now) for faster downloading
3. can specify directories for downloading episodes.
4. can downloading in non-linux partitions (e.g. - ntfs, fat32,exfat etc).. by allocating space first...
5. can also download torrents... if in the future we add support for torrent anime websites (subplease.org)

And many more...[read here](https://aria2.github.io/manual/en/html/aria2c.html)

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -356,10 +356,11 @@ open_episode () {
 		mkdir -p "$download_dir"
 		printf "Downloading episode $episode ...\n"
 		printf "%s\n" "$video_url"
+		printf "%s\n" "$play_link"
 		# add 0 padding to the episode name
 		episode=$(printf "%03d" $episode)
 		{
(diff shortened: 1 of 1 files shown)


ura43:
Can confirm working.
Parallel downloading is unsupported by gogo you say?

CoolnsX:
What I mean by parallel downloading is that.. it splits the file into parts and then download them properly...

And right now gogoanime are using fast servers... Like really fast.. I get 12-13 MBps... My whole internet speed..

CoolnsX:
Also pls add... aria2c as dependency requirement in readme as well as in script.... I forgot to add it in readme..

ura43 (approved):

ura43:
as discussed in #255

Issue #255, opened 2022-01-01
Can we make ani-cli more minimal again?


ura43:
Bold move, I know, but the makefile has definitely caused way more harm than good.
ani-cli is a simple shellscript, that's the beauty of it, symlinking it to /usr/local/bin just for it to be in the path is not worth the countless issues that have popped up and continue to pop up around it.

It was originally meant to make installation easier, but the following issues tell a different tale: #247  #246 #242  #241 #205 #178 #164  #159 

I'm aware that this breaks the current automatic update checking, but that one is janky at best as well and doesn't lead to a good user experience. I like to work in a "move fast, break things" kinda way and would hate to have a user who has no issues using ani-cli update only to have something break for them. Tho a "If you have bugs, update first" disclaimer in the contributing.md wouldn't hurt.

The icing on the cake: we get rid of the make and git dependency

ura43:
@RaynardGerraldo @Dink4n

RaynardGerraldo:
> Bold move, I know, but the makefile has definitely cause way more harm than good.
> ani-cli is a simple shellscript, thats the beauty of it, symlinking it to /usr/local/bin just for it to be in the path is not worth the countless issues that have popped up and continue to pop up around it.
> 
> It was originally meant to make installation easier, but the following issues tell a different tale: #247  #246 #242  #241 #205 #178 #164  #159 
> 
> I'm aware that this breaks the current automatic update checking, but that one is janky at best as well and doesn't lead to a good user experience. I like to work in a "move fast, break things" kinda way and would hate to have a user who has no issues using ani-cli update only to have something break for them. Tho a "If you have bugs, update first" disclaimer in the contributing.md wouldn't hurt.
> 
> The icing on the cake: we get rid of the make and git dependency

I agree, its also not helping considering this is more of a mac bug and ani cli cant even build in m1 chip.

So lets say if the makefile is removed,we'll be running the script with a simple ./ani-cli right?

And for the automatic update checking part,at first i kinda like the feature ,but seeing what you just said about user updating ani-cli only for it to break makes me think that maybe we could keep the feature but make it optional? But then again user would have to find out that it broke for them after they updated so i think its best to just remove it for now at least until we find some sort of solution 

And tbh I havent run makefile with ani-cli since the first fix I made,its so annoying when I thought my fix didnt work and turns out I didnt do sudo make yet,just stopped using it after that and ran it with ./ani-cli😂

ura43:
[...]

PR #258, opened 2022-01-01
Makefile for Mac and Linux

KermitPurple:
It seems to me that `sed` and `ln` have slight differences on Mac and Linux.
On Mac, `sed`'s `-i` flag requires an argument.
On Mac, `ln` does not have a `-r` flag.
Because of this, I used `uname` to check which operating system is running, and used different flags in the case of Mac vs Linux.

Additionally, I put `@` in front of the comments to make it so `make` does not print them.


ura43:
While this PR is likely to fix our Problems, it does so by making things more complicated.
The consensus from #255 is that we want to move back to a simpler makefile

PR #259, opened 2022-01-02
Stop current episode from continuously playing when switching between episodes.

Kungger-git:
This short statement is pretty convenient in my opinion when trying to switch between episodes. I used `pgrep` to check if the process of `mpv` or `vlc` is still running.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -338,7 +338,7 @@ open_episode () {
 					rm "${logfile}.new"
 				else
                                         if ps "$PID" &> /dev/null; then
-						kill "$PID"
+					        kill "$PID"
 					fi
 					nohup $player_fn "$play_link" > /dev/null 2>&1 &
 					PID="$!"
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -337,6 +337,9 @@ open_episode () {
 					printf "${c_red}\nCannot reach servers!"
 					rm "${logfile}.new"
 				else
+                                        if pgrep mpv; then
+                                            pkill mpv
+                                        fi
 					nohup $player_fn "$play_link" > /dev/null 2>&1 &
```
Dink4n, line 343: Instead of using `pgrep` you can get the PID of last ran program with
```
PID=$!
```
So, just before running `mpv` we can kill the previous instance like this
```
kill $PID
```

It will look something like this
```
kill $PID
nohup stuff
PID=$!
```

Kungger-git: I figured it out on how to implement that, though having `kill $PID` without having any defined process will return an error. Instead I used this:
```
if ps "$PID" &> /dev/null; then
    kill "$PID"
fi
```
so that it checks if a previous instance is created.

Dink4n: Actually, we don't need to check for a previous instance. We can make `kill` silent
```
kill "$PID" >/dev/null 2>&1
```

ura43 (approved):

PR #256, opened 2022-01-01
Support for more players and removal of ampersands

tennyson-mccalla:
This allowed iina to work for me

PR #260, opened 2022-01-02
Allow the user to choose his player

TheCactusVert:
- Allow the user to choose his player.
- Adding Celluloid as allowed player.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -33,7 +33,7 @@ help_text () {
 	 -q	 set video quality (best/worst/360/480/720/..)
 	 -k  on keypress navigation (previous/next/replay/quit episode)
 	 --dub	 play the dub version if present
-	 -v  use VLC as the media player
+	 --player=<PLAYER> use <PLAYER> as the media player
    
 	add 'h' on beginning for episodes like '6.5' -> 'h6'
 	EOF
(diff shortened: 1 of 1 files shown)


ura43:
@RaynardGerraldo is already working with @tennyson-mccalla on implementing this.
If you want to review their implementation, check out #256

django85:
This would be great.
E.g. I'm running OSMC and using the kodi-send command to play the video in ani-cli:
kodi-send --action="PlayMedia($play_link)"

A generic solution by using a config.ini file would be neat, something like this (in case of kodi):

[player]
cmd=kodi-send
param_string=--action="PlayMedia($play_link)"

ura43:
We 100% won't use a config file @django85

RaynardGerraldo:
> This would be great. E.g. I'm running OSMC and using the kodi-send command to play the video in ani-cli: kodi-send --action="PlayMedia($play_link)"
> 
> A generic solution by using a config.ini file would be neat, something like this (in case of kodi):
> 
> [player] cmd=kodi-send param_string=--action="PlayMedia($play_link)"

that would be too much of a hassle to implement, the plan for the support of multiple video players is basically match wildcard options and will only work with media player that can execute videos after its command, for ex: vlc <link>

PR #272, opened 2022-01-03
Add ani-gui

SingularisArt:
Hello there.

I created another script that is a `gui` instead of a `cli based interface`. Unfortunately, it only works on Linux operating systems because you need the program [rofi](https://github.com/davatorium/rofi).


ura43:
Uhm, how about you make your own repo for that?

PR #280, opened 2022-01-03
A rewrite in Crystal lang, better comments, communication with mpv's JSON IPC

lncndev:
Firstly, I know this is a breaking change and that some may not agree with it, but I feel like it is beneficial for the future of the project. Also, this is a few days old so not all new functionality has been ported over yet. But anyways, my reasoning:
- Bash is great, but creating anything complex in a single file will end up becoming a mess
- Compiled code is much faster
- Commenting and documentation is much easier in Crystal
- ani-cli's bloat issues can be solved by just not compiling extra features only some people will actually need or find useful

---

### Resources
[Crystal language reference](https://crystal-lang.org/reference)
[Crystal API docs](https://crystal-lang.org/api)

[crystal-term shards](https://github.com/crystal-term) (used for interactive prompt and screen dimensions)
[mpv JSON IPC docs](https://mpv.io/manual/stable/#json-ipc)

---

I'm happy to answer any further questions :)


ura43:
What the hell?
Make your own repo

ura43:
On Shellscripts:
Yes, it's not the ideal language for a scraper. 
The ideal language would be python.
[animdl](https://github.com/justfoolingaround/animdl) is a superior program in all technical regard.
But keeping this as a shellscript will allow our target audience, linux enthusiasts, to easily adapt the script to their needs.

On compiled languages:
Yes they run faster, but any improvement elsewhere than the bottleneck is no improvement.
This is a webscraping script. 
A single web request to gogo's servers will take longer than all our helper functions combined, and easily by an order of magnitude.

On documentation:
Our current documentation is non-existant, which sucks. 
Documentation is planned tho.

On bloat:
A patch set, similar to how suckless is doing it, seems more appealing to me personally, but I do see the appeal of your approach

ura43:
There is definitely a market for a well documented crystal gogo scraper.
I welcome competition and wish us and animdl weren't the only maintained downloaders

ura43:
If you want to maintain the crystal rewritten fork, that would be nice to see.
But either way I'll close this PR now since it's getting kinda stale and will not be merged anyway

PR #283, opened 2022-01-04
the link from the script

4cecoder:


ura43:
Why? The old domain still redirects to gogoanime and they have a history of changing their url on a regular basis

4cecoder:
because we can update it whenever they do just like the script

PR #290, opened 2022-01-04
Replace curl with aria2

CoolnsX:
As we know that, gogoanime enabled referer link for downloading... we now get reduced speed.. like 400-600 KBps... Now we can use parallel downloading of aria2... to get above 1-2 MBps speed..
@Dink4n @ura43 you can try it with or without aria2.. by checking any episode..

note:- some download links ends at "==" or "=" .. they are little slower than the links ending at "alphabet" or "numeric"... In my testing...

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ A cli to browse and watch anime. This tool scrapes the site [gogoanime](https://
 grep
 curl
 sed
-ffmpeg
+aria2
 git
 ```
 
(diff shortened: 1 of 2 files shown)


ura43:
I'll do some benchmarking later today, but I certainly found that download speeds were inconsistent throughout the past few days.
Sometimes I got my full bandwith, other times it took 10s to buffer 1s of video

ura43 (approved):
Nice, you even updated the readme

PR #292, opened 2022-01-04
Test with shellcheck

ura43:


ura43:
Our printf usage is still questionable, but this is a big first step in the right direction

ura43:
Actually, I have some first ideas on how to fix that

ura43:
> Actually, I have some first ideas on how to fix that

They did not work

PR #299, opened 2022-01-05
Feature add: update script

CarlosLoboxyz:
This snipet will allow you to update your ani-cli instance automatically
for _**future releases only**_

It's pretty handy for people who have their ani-cli inside `.local/bin` or something like that

Edit: Someone could say
>You could just symlink ani-cli from the git repo

But i think this make the script more usefull for people who only want to update without messing with git

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ A cli to browse and watch anime. This tool scrapes the site [gogoanime](https://
   ```
     ani-cli [-kv] [--dub] [-q <quality>] [-d | -p <download_dir>] [<query>]
     ani-cli [-kv] [--dub] [-q <quality>] -u | -n | -H
-    ani-cli -h | -D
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,6 +46,8 @@ A cli to browse and watch anime. This tool scrapes the site [gogoanime](https://
 ```
 grep
 curl
+diff
```
Dink4n, line 49: Can you move these dependencies to optional

Dink4n (approved):

ura43:
We had our fun with symlinking and git update checking. Most fragile thing ever. This is better for sure

CoolnsX:
Yup.. looks pretty solid to me...

ura43 (approved):

ura43:
Good work, merging now

Issue #334, opened 2022-01-14
VLC tries to load all files from directory if $video_url is null


galloway:
When trying to play an episode through ani-cli and the $video_url is empty, the command that launches the player  loads all files in the working directory:

`nohup vlc --http-referrer=https://sbplay2.com  > /dev/null 2>&1 &`

~~Please note that running the above command directly from the terminal does not seem to have the same effect on my system.~~

EDIT:  
`nohup vlc --http-referrer=https://sbplay2.com "" > /dev/null 2>&1 &`  
instead recreates the issue.

galloway:
I've created a simple PR #335 to check for this and error out instead.

galloway:
I just realized this is actually a duplicate of #312.

Issue #323, opened 2022-01-08
After gathering data for an episode VLC launches but closes right after


Chiroyasumi:
Everything works seamlessly until showing the episode. I use windows and VLC, there is a video attached

https://user-images.githubusercontent.com/51413351/148646919-7e6b0d15-2d48-4103-9de9-3a9585456897.mp4

JRCdev:
Seems to be an issue with getting the video_url from dpage_link

ura43:
As JRCdev said, this might just be vlcs symptoms of #312

Chiroyasumi:
> As JRCdev said, this might just be vlcs symptoms of #312

Ooh okay, sorry for the duplicate then

ura43:
No worries. Keep an eye out for a fix, it might take a few days this time

PR #335, opened 2022-01-14
Added a simple check for a null $video_url

galloway:
Should be a simple fix for issue #334, #323, etc.


ura43 (approved):

ura43:
Better merge now before the major restructuring takes place

PR #338, opened 2022-01-14
switched to vidcdn

CoolnsX:
@ura43  please check it...


ura43 (changes requested):
Would it be hard to get rid of the jq requirement?

CoolnsX:
> Would it be hard to get rid of the jq requirement?

yup... as it will make the scripting easier.. because hard coding everything is time consuming results in complexed script..

ura43:
Currently discussing it with some users on the discord

ura43:
But avoiding additional dependencies seems like a good default strategy

ura43 (approved):

PR #346, opened 2022-01-15
Incorporated vidcdn updates

JRCdev:
Improvements added:
 - Added ajax code matching current linux master
 - updated download code to use aria2c if user's got it on their path but if they don't it'll use curl
 - Filters out .vtt files that sometimes show up in the list of download links

Limitations:
 From my limited testing it still gets the odd 403 errors with some series and settings but not others:
 - Fine for the FLCL dubs but not subs
 - Fine for Serial Experiments Lain but not Legend of the Galactic Heroes
But with this it once again plays _many_ things as opposed to _no_ things. If these bugs also exist in linux version this'll get windows users back to that level of access.

### DIFF
--- a/ani-cli-win
+++ b/ani-cli-win
@@ -89,6 +89,7 @@ decrypt_link() {
 
     #send the request to the ajax url
     curl -s -H 'x-requested-with:XMLHttpRequest' "$ajax_url" -d "id=$ajax" -d "time=69420691337800813569" \
+    		| jq '.' \
 		| tr '"' '\n' \
 		| sed -n -E 's/.*cdn.*/\0/p' \
 		| sed 's/\\//g' \
@@ -229,9 +230,7 @@ open_episode () {
 	printf "Getting data for episode %d\n" $episode
 
 	dpage_link=$(get_dpage_link "$anime_id" "$episode")
-	# printf "%s\n" "$dpage_link"
 	video_url=$(get_links "$dpage_link")
-	# printf "%s\n" "$video_url"
 
 	case $video_url in
 		*streamtape*)
--- a/ani-cli-win
+++ b/ani-cli-win
@@ -25,7 +25,7 @@ help_text () {
 	 -h	 show this help text
 	 -d	 download episode
 	 -H	 continue where you left off
-	 -q	 set video quality (best/worst/360/480/720/..)
+	 -q	 set video quality (best|worst|360|480|720|1080)
 	EOF
 }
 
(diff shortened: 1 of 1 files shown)


ani-cli-win
```diff
@@ -213,7 +229,9 @@ open_episode () {
 	printf "Getting data for episode %d\n" $episode
 
 	dpage_link=$(get_dpage_link "$anime_id" "$episode")
+	# printf "%s\n" "$dpage_link"
 	video_url=$(get_links "$dpage_link")
+	# printf "%s\n" "$video_url"
 
 	case $video_url in
```
CoolnsX, line 236: Can u remove from line 236 to 247.. these lines are useless.. it doesn't do anything..

CoolnsX: Also from line 92 to 95.. cant't we use jq??..
Because these line will be helpful in new anime.. but in older anime.. it will produce useless link..
I know adding jq as dependency will be a problem in windows.. but if the script is not up to the task then it will make the windows user ditch up the idea of ani-cli

JRCdev: Oh, sure!
Also, I could make it so it only uses jq if the user has jq installed, would that be an alright compromise? I'm not sure what the best UX decision is.

ani-cli-win
```diff
@@ -72,32 +72,48 @@ get_dpage_link() {
...
 
 		*)
-			video_link=$(printf '%s' "$video_links" | grep -i "${quality}p")
+			video_link=$(printf '%s' "$video_links" | grep -i "${quality}p" | head -n 1)
 			if [ -z "$video_link" ]; then
-				printf "$c_red%s$c_reset\n" "Current video quality is not available (defaulting to highest quality)" >&2
+				err "Current video quality is not available (defaulting to best quality)"
 				quality=best
-				video_link=$(printf '%s' "$video_links" | tail -n 1)
+				video_link=$(printf '%s' "$video_links" | head -n 4 | tail -n 1)
```
ura43, line 116: Why does this need changing? Is the video link now always in line 4, with padding below sometimes?

ani-cli-win
```diff
@@ -234,15 +252,16 @@ open_episode () {
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new" && mv "${logfile}.new" "$logfile"
 
-		$player_fn "$video_url" "vlc://quit" --http-referrer="$dpage_link"  --http-user-agent="Mozilla/5.0" --adaptive-use-access  >/dev/null 2>&1
+		nohup "$player_fn" --http-referrer="$dpage_link" --http-user-agent="Mozilla/5.0" --adaptive-use-access "$video_url" "vlc://quit" > /dev/null 2>&1 &
```
ura43, line 255: This is part of #256 but could you remove the nohup ampersands while you're at it?
They are confirmed not necessary on all supported platforms

ani-cli-win
```diff
@@ -234,15 +252,16 @@ open_episode () {
...
 		printf "Downloading episode $episode ...\n"
 		printf "%s\n" "$video_url"
 		# add 0 padding to the episode name
 		episode=$(printf "%03d" $episode)
 		{
-			curl -4 -L -# -C - "$video_url" -G -e 'https://streamani.io/' \
-				-o "${anime_id}-${episode}.mp4" "$video_url" >/dev/null 2>&1 &&
+			(command -v aria2c &&
+			aria2c -x 16 -s 16 --referer="$dpage_link" "$video_url" --dir="$download_dir" -o "${anime_id}-${episode}.mp4" --download-result=hide ||
+			curl -4 -L -o "${anime_id}-${episode}.mp4" "$video_url" >/dev/null 2>&1 ) &&
```
ura43, line 264: I love this syntax. 
`test && optional feature || default behaviour` should be more common in our codebase

ura43 (changes requested):
Good work so far

ani-cli-win
ura43: Oh, this is just backporting fixes to the windows branch. My bad

ani-cli-win
ura43: Never mind, since this is a backport...

ura43 (approved):

ura43:
I didn't spot this was backporting fixes to the windows branch. Please ignore my requested changes from above

ani-cli-win
ura43: No. Definitely don't support both. If jq is needed elsewhere in the windows branch already, you might as well port the parsing to jq, but if windows got spared with that dependency then don't change it.
Installing software on windows is more of a hassle and it is safe to assume our average windows user to be less technically literate than our average linux user.
Removing the comments does make sense tho

JRCdev:
I removed the comments and added jq. The latter didn't add or remove functionality afaik, but I've discovered that more of my test cases work when asking for the worst quality explicitly.

Issue #348, opened 2022-01-16
Introduce versioning


ura43:
I propose we either use x.x.x or x.x
First version would be coupled to a milestone and either 1.0 or 1.0.0
Thoughts?

ura43:
https://semver.org/

major minor seems more sensible, since we break backwards compatibility all the time

PR #349, opened 2022-01-16
Add version number

3ricsonn:
I added a version number for reference to see which practical version is used in case of a bug, so misconception are more unlikely to accrue

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # version umber
-VERSION="1.0"
+VERSION="1.0.0"
 
 
 # dependencies: grep sed curl video_player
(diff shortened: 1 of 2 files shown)


3ricsonn:
Would be a suggestion to #348

ani-cli
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+# version umber
+VERSION="1.0"
```
ura43, line 4: Actually please set 1.0.0 since three numbers seem to be best practice here on github

ura43 (changes requested):

ani-cli
ura43: actually its such a small change I can do that myself

ura43 (approved):

PR #352, opened 2022-01-17
Improve UX and slim down

ura43:
- [x] `-H` is just a worse form of `-u` **(remove)**
- [x] `--dub` prompts you for selecting an anime, even if only one version was found **(remove)**
- [x] `-h` needs another sorting *(polish)*
~`-k` should become more consistent and maybe even default behaviour *(rewrite)*~
~`-p` should really be an optional arg for `-d` **(remove)**~
~`-v` should allow choice in any non-default video player as optarg, with vlc staying the default alternative.~

As always this tasklist is subject to change

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -11,6 +11,8 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Run ShellCheck
-      uses: ludeeus/action-shellcheck@v1.1.0
+      uses: ludeeus/action-shellcheck@master
+      with:
+        version: v0.7.2
       env:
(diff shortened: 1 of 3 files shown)


ura43:
I've been made aware that we could just delete the history entry for finished series. I'll make a poll about that, but it seems like a good idea and could replace the sizeable helper function currently employed by -u

ura43:
-k does not appear sensible to integrate on a larger scale since searches can give more than 10 results and most anime have more than 10 episodes

ura43:
-v will be implemented by the other PR

ura43:
optional parameters are cancer

PR #371, opened 2022-01-21
Code cleanup

Derisis13:
Putting everything I can into functions and organising them into logical groups. Also adding & correcting comments


ura43:
Please pull in the latest changes or close and reopen another.
Also the search_unwatched function is needed for `-c`

Derisis13:
Working on it!
Also cleanup suggestions are welcome

Derisis13:
Some of my ideas: 
Moving the aria2c checking into dep_ch (but retaining the correct package name)
One function to display dual-colour menus
Modifying check_input and reusing it in anime_selection

ura43:
So far things are looking good. The additional section comments really help make the code easier to digest.

> Moving the aria2c checking into dep_ch (but retaining the correct package name)

Sure, why not.

> One function to display dual-color menus

I think we should do more than that: adding some form of printf wrapper function and then just calling it from wherever we currently use printf would be a good idea.
That way we can ensure style consistency and have an oportunity to finally use printf in a shellcheck compliant way.
That could take a bit, so take your time with that

> Modifying check_input and reusing it in anime_selection

I don't think we should do that before we have implemented ranges for streaming

Derisis13:
I don't know what is going on with the spellcheck: It might be tripping on `CONTRIBUTING.md` because the exit code is OK for me. On the other hand the script doesn't open `mpv` for some reason, I have to debug that.
I have a few more ideas for `printf` wrappers (eg. menu-lines, colored user input, ect.). Once I run out of ideas and the script is working I'll mark the commit ready.

Questions:
Should line 79 be `err` or `inf`?
Line 507 greatly suffered in legibility from being converted to `inf`. Extra colours could be added with their variables, but that would make the code hard to read again. (I think the goal with separating the `printf`-s into function calls is more consistent look and the code being easier to understand.) Should I work on it more or leave it as is?
Are lines 354, 355 and 358 necessary? They are super nice for debugging, but an eyesore for the end user.
Is there a smart way to output a newline (like `tput reset` for clearing the screen)? I wish to avoid `echo ` if possible, and it is much needed.

ura43:
Beware the [SC2015](https://github.com/koalaman/shellcheck/wiki/SC2015)
I like the err, inf, die paradigm.
More in depth review coming

ani-cli
```diff
@@ -308,13 +346,13 @@ open_episode () {
...
-	printf "Getting data for episode %s\n" "$episode"
+	inf "Getting data for episode $episode"
 
 	dpage_link=$(get_dpage_link "$anime_id" "$episode")
 	video_url=$(get_video_quality "$dpage_link")
 
-	printf "%s\n" "$dpage_link"
-	printf "%s\n" "$video_url"
+	inf "$dpage_link"
+	inf "$video_url"
```
ura43, line 355: Line 354, 355 and 358 aren't necessary

ani-cli
```diff
@@ -44,32 +49,47 @@ help_text () {
...
 
+dep_ch () {
+	# checks if programs are present 
+	for dep; do
+		if ! command -v "$dep" >/dev/null ; then
+			err "Program \"$dep\" not found. Please install it."
+			#aria2c is in the package aria2
+			if [ "$dep" = "aria2c" ]; then
+				# echo, inf or err?
+				echo "To install aria2c, Type <your_package_manager> aria2"
```
ura43, line 79: Should be err / die. die makes most sense since that gets rid of the `exit 1` on line 81

ani-cli
```diff
@@ -465,8 +498,13 @@ check_input
...
 open_selection
 
+
+########
+# Loop #
+########
+
 while :; do
-	printf "\n${c_green}Currently playing %s episode ${c_cyan}%d/%d\n" "$selection_id" "$episode" "$last_ep_number"
+	inf "Currently playing $selection_id episode $episode/$last_ep_number"
```
ura43, line 507: I think the sacrifice in UI prettiness is worth the gain in uniformity. You can leave this as is

ura43 (approved):

Derisis13:
> Beware the [SC2015](https://github.com/koalaman/shellcheck/wiki/SC2015) 
Not even my code, but I'll fix it. This is a cleanup after all

ani-cli
Derisis13: The `exit 1` is outside of the `if`-`fi` branch, so using die wouldn't replace it. Using die on line 75 is a bad idea too, since it'd cause the following lines to be never executed.
On the other hand instead of `exit 1` an empty `die` would make sense.

ura43:
Consider making a prompt helper function that doesn't print a new line after itself.
Just tested it and it would be kind of a step back in terms of UI.
The perfect color for a prompt would be c_blue

Derisis13:
> Consider making a prompt helper function that doesn't print a new line after itself. Just tested it and it would be kind of a step back in terms of UI. The perfect color for a prompt would be c_blue

Ok, I'll do it. I think the user should still write in the default colour, but having everything in one line is nice.

Derisis13:
The local thing is easy to remove, but the only POSIX-compliant substitute for `printf -v` I found is `eval` which I read is best avoided.
I suppose eval is neccesary to stay compliant with all platforms.

ura43:
There should be some way to avoid eval. If nescessary, make the previously local variables non-local. Doesn't matter if we use up 1 KB more ram if it saves us from eval hell.

ani-cli
```diff
@@ -93,6 +115,32 @@ search_eps () {
...
+		then
+			unwatched_anime="$unwatched_anime$anime_id\n"
+		fi
+	done <<-EOF
+	$search_results
+	EOF
+	if [ -z "$unwatched_anime" ]; then
+		die "No unwatched episodes"
+	fi
+	printf "%s""$unwatched_anime"
```
pystardust, line 136: there would be a space between the format and args here

Derisis13: Nice find!

Derisis13:
`eval` is inevitable... `typeset` isn't POSIX, `declare` isn't POSIX, `printf -v` isn't POSIX, `read -r <<<` isn't POSIX. I could find no more solutions to this indirection problem....

ani-cli
```diff
@@ -150,54 +199,69 @@ get_video_quality() {
...
-[ -f "$logfile" ] || : > "$logfile"
+prompt () {
+        # prompts the user with message in $1 and saves the input to the variables in $2 and $3 
+        printf "$c_blue%s: $c_reset" "$1"
+        read -r input1 input2
+        # indirectly assigning the input to the variable in $2 and in $3 if present
+        if [ -z "$3" ]; then
+                eval "$2=\$input1"
+        else
+                eval "$2=\$input1"
```
pystardust: one alternative here is to remove eval and read some variable called say `REPLY` and not have `$2` at all so it would look like. So that this variable would be accessible outside the function like so
```
prompt "Select episode: "
episode=$REPLY
```

ani-cli
```diff
@@ -150,54 +199,69 @@ get_video_quality() {
...
+                eval "$3=\$input2"
+        fi
+}
 
-#####################
-## Anime selection ##
-#####################
+menu_line_even () {
+	# displays an even (cyan) line of a menu line with $2 as an indicator in [] and $1 as the option
+	printf "$c_blue[$c_cyan%c$c_blue] $c_cyan%s$c_reset%s\n" "$2" "$1"
```
pystardust, line 230: this is a lil bit of a shell check thing. The shell check confuses this with arrays in bash since there is a `$variable[...]` variable and bracket. So to avoid this you could do `${variable}[`

ani-cli
Derisis13: And that would be even less confusing, even though we need 2 replies because of episode ranges

ani-cli
Derisis13: Thanks, I'm new to shellcheck so I don't know the ins and outs of it.

ani-cli
```diff
@@ -150,54 +199,67 @@ get_video_quality() {
...
 
-# create history file
-[ -f "$logfile" ] || : > "$logfile"
+prompt () {
+        # prompts the user with message in $1 and saves the input to the variables in $2 and $3 
+        printf "$c_blue%s: $c_reset" "$1"
+        read -r input1 input2
+        # indirectly assigning the input to the variable in $2 and in $3 if present
+	eval "$2=$input1"
+        if [ "$3" ]; then
```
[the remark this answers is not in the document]
Derisis13: Done, I just didn't push it to github

ani-cli
ura43: > this is a lil bit of a shell check thing. The shell check confuses this with arrays in bash since there is a `$variable[...]` variable and bracket. So to avoid this you could do `${variable}[`

100% correct. Do you think I should raise an issue for that on their repo?

Derisis13: It backfired, because it put the `{}`-s into the string...

Derisis13:
At this point everything I can think of regarding refactoring is done. The only exception is that `open_episode` does way more things than "opening an episode" without function calls, but I don't understand it well enough to touch it.
`eval` is gone, shellcheck passed so I'll mark the PR as ready.

ani-cli
```diff
@@ -465,58 +509,53 @@ check_input
...
-		printf "$c_blue[${c_cyan}%s$c_blue] $c_yellow%s$c_reset\n" "n" "next episode"
+		menu_line_odd "next episode" "n"
 	fi
 	if [ "$episode" -ne "1" ]; then
-		printf "$c_blue[${c_cyan}%s$c_blue] $c_magenta%s$c_reset\n" "p" "previous episode"
+		menu_line_even "previous episode" "p"
 	fi
 	if [ "$last_ep_number" -ne "1" ]; then
-		printf "$c_blue[${c_cyan}%s$c_blue] $c_yellow%s$c_reset\n" "s" "select episode"
+		menu_line_odd "select episode" "s"
```
ura43, line 526: Is there a way to avoid hardcoding alternation?
image

ani-cli
```diff
@@ -44,32 +49,49 @@ help_text () {
...
-		printf "$c_green%s$c_reset\n" "Script is up to date :)"
+		inf "Script is up to date :)"
 	else
-		printf '%s\n' "$update" | patch "$0" - && printf "$c_green%s$c_reset\n" "Script has been updated" ||
-            printf "$c_red%s$c_reset\n" "Can't update for some reason!"
+		if printf '%s\n' "$update" | patch "$0" - ; then
+			inf "Script has been updated"
+		else
+            		die "Can't update for some reason!"
+            	fi
```
ura43, line 70: Weird indentation

ani-cli
```diff
@@ -150,54 +201,63 @@ get_video_quality() {
...
+        printf "$c_blue%s$c_magenta%s$c_blue: $c_reset" "$1" "$2"
+        read -r REPLY REPLY2
+}
 
-#####################
-## Anime selection ##
-#####################
+menu_line_even () {
+	# displays an even (cyan) line of a menu line with $2 as an indicator in [] and $1 as the option
+	printf "$c_blue[$c_cyan%c$c_blue] $c_cyan%s$c_reset%s\n" "$2" "$1"
```
ura43, line 227: What's the status on the shellcheck here?
Considering `printf` doesn't print a newline by default, couldn't you just make it two `printf`s to dodge the false positive?

ani-cli
```diff
@@ -150,54 +201,63 @@ get_video_quality() {
...
+}
+
+menu_line_odd() {
+	# displays an odd (yellow) line of a menu line with $2 as an indicator in [] and $1 as the option
+	printf "$c_blue[$c_yellow%c$c_blue] $c_yellow%s$c_reset%s\n" "$2" "$1"
+}
```
ura43, line 233: Could you explain your coloring?

ura43 (approved):
Can confirm working, at least from testing the default usecase of streaming an episode without additional parameters

ani-cli
Derisis13: Green is information, magenta is auxiliary information. -> inf
Red is warning/error/attention. -> err, die, menu_line_strong
Blue is when asking for user input and for the brackets in the menus (for the lack of better). -> prompt, menu_line*
Yellow and cyan are the alternating colours in the menu and they extend to the options in the brackets -> menu_line_even, menu_line_odd
For user input, colour is always reset.

If you think other colours would go together better I'll implement it. I have deuteranopia so leaving the colours up to me is not a good idea...

ani-cli
Derisis13: I can do that, but it'd require an if-else statement at every option. I don't think it's worth it.

ani-cli
Derisis13: I see that it passed the shellcheck in its current form. Really similar code is in the master branch. Why would I change it?

ura43: It didn't, we just ignore SC1087 and SC2059 currently

ura43:
On lines 168 and 173 inside the decrypt_link function, can we get rid of piping into the other stuff? Those are the last two instances of SC2059

ani-cli
Derisis13: Then I'll split it

Derisis13:
> On lines 168 and 173 inside the decrypt_link function, can we get rid of piping printf into the other stuff? Those are the last two instances of SC2059

Their output is not `stdout` nor `stderr`. They can be substituted with echo though. Would that do it?

Derisis13:
> > On lines 168 and 173 inside the decrypt_link function, can we get rid of piping printf into the other stuff? Those are the last two instances of SC2059
> 
> Their output is not `stdout` nor `stderr`. They can be substituted with echo though. Would that do it?

Oh, adding "%s" would pass the spellcheck. I'll do that and not convert to echo.

ura43:
I'd prefer just passing them as the final argument to the next command down the pipeline (cut and openssl respectively

There is another part where there are two useless echos on line 297 and 300

Or do you think that hurts readability?

Derisis13:
I see no problem in that, in fact I'll do it that way.

Derisis13:
I thought I knew what you meant, but I had to realise I don't. Can you explain in more detail what 
>just passing them as the final argument to the next command

 means? `cut [...] $var` doesn't work for me and `cut [...] <<< $var` is not POSIX compliant...

ani-cli
ura43: Fair enough, if no elegant solution occurs to us over night we'll leave it like that

ura43:
> I thought I knew what you meant, but I had to realise I don't. Can you explain in more detail what
> 
> > just passing them as the final argument to the next command
> 
> means? `cut [...] $var` doesn't work for me and `cut [...] <<< $var` is not POSIX compliant...

At least for openssl I can't find how to make it work without somehow piping it in.
For the 3 cuts it should work fine

ura43:
But still, at that point we might as well establish `echo` to be the default piping slave

Derisis13:
So yeah, SC[2059] was ignored for a reason.
I also tried to make alternating colors not be hardcoded, but arrays aren't shell compliant. I don't feel confident enough in my shell skills to do it without arrays, so I'll just open an issue for now and leave it as is.

ura43:
Its fine, you have done plenty. I'll take another look at it tomorrow and then probably merge before lunch

Derisis13:
Critical bug found: n and p don't quite work, I'll have to debug it. The behaviour is very cursed: 
image

ura43:
Tested successfully, the few remaining rough edges should be done in another PR if at all

ura43:
Maybe placebo, but I feel like this PR made the script faster

PR #375, opened 2022-01-22
Some bugfixes and a new feature

71zenith:
1) Made so that if the script fails to update for some reason, it
       will tell the user
    2) Fixed a bug which won't print version if an argument isn't
       provided
    3) Made so that the program will only try to get the url if
       neccessary(E.G. not when the user prints the help menu)
     ** EXTRA DEPENDENCY ADDED(TAC)**
    4) Made so that history will be shown reverse so the last seen anime
       will be at the top
    5) Made so that if an anime is watched with it already being on the
       history, the script will delete the line and append to the end

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,10 +1,9 @@
 #!/bin/sh
 
 # version number
-VERSION="1.1.0"
+VERSION="1.1.2"
 
 logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
-base_url=$(curl -s -L -o /dev/null -w "%{url_effective}\n" https://gogoanime.cm)
 
 c_red="\033[1;31m"
 c_green="\033[1;32m"
@@ -64,8 +63,8 @@ update_script () {
 	if [ -z "$update" ]; then
 		printf "$c_green%s$c_reset\n" "Script is up to date :)"
 	else
-		printf '%s\n' "$update" | patch "$0" -
-		printf "$c_green%s$c_reset\n" "Script has been updated"
+		printf '%s\n' "$update" | patch "$0" - && printf "$c_green%s$c_reset\n" "Script has been updated" ||
+            printf "$c_red%s$c_reset\n" "Can't update for some reason!"
 	fi
 }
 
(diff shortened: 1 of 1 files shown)


71zenith:
This is my first time making any pull request. So please don't mind if you see any mistake. Do tell me though

ani-cli
```diff
@@ -64,8 +63,8 @@ update_script () {
 	if [ -z "$update" ]; then
 		printf "$c_green%s$c_reset\n" "Script is up to date :)"
 	else
-		printf '%s\n' "$update" | patch "$0" -
-		printf "$c_green%s$c_reset\n" "Script has been updated"
+		printf '%s\n' "$update" | patch "$0" - && printf "$c_green%s$c_reset\n" "Script has been updated" ||
+            printf "$c_red%s$c_reset\n" "Can't update for some reason!"
```
ura43, line 67: Sure, that seems like a good idea

ani-cli
```diff
@@ -377,7 +376,7 @@ quality=best
 scrape=query
 download_dir="."
 
-while getopts 'viq:dp:chDUV:' OPT; do
+while getopts 'viq:dp:chDUV' OPT; do
```
ura43, line 379: I thought we needed that to catch the anime query, but turns out we don't. Nice catch.

ani-cli
```diff
@@ -444,14 +443,16 @@ shift $((OPTIND - 1))
...
+        base_url=$(curl -s -L -o /dev/null -w "%{url_effective}\n" https://gogoanime.cm)
 		get_search_query "$*"
 		search_results=$(search_anime "$query")
 		[ -z "$search_results" ] && die "No search results found"
 		anime_selection "$search_results"
 		episode_selection
 		;;
 	history)
-		search_results=$(sed -n -E 's/\t[0-9]*//p' "$logfile")
+        base_url=$(curl -s -L -o /dev/null -w "%{url_effective}\n" https://gogoanime.cm)
```
ura43, line 454: Please move the base_url assignment before the case statement

ura43 (changes requested):
While this PR has some good things, I don't see tac being useful enough to warrant its inclusion

ura43:
Oh, also bump the version to 1.1.2

ani-cli
71zenith: Did everything

ura43 (approved):

Issue #377, opened 2022-01-22
App crashes when choosing invalid option in main episode view


YairZiv5951:
I'm using 2 languages on my keyboard, sometimes I don't notice and write in the main language I use.
If possible, can you make it so that in the switch case statement at the end of the file, when choosing something outside of the available choices it print the error "invalid token" but not kill the app? Just let you choose another?

Thanks!

CoolnsX:
It's possible.. but line of code will increase..

71zenith:
That would be a long rewrite tho

ura43:
Fixed in #384 
Was literally a 2 line addition btw

PR #384, opened 2022-01-24
fixed app closed on invalid option in ep interface

CoolnsX:
fixed the issue #377 , along with interface not showing for those having nohup version > 8.3


ura43:
Its nohup version 8.30, as in 8.29++

ura43:
Can confirm working. Added the tput reset so that there is only one instance of the menu on the screen at a time -> better UI

PR #386, opened 2022-01-24
Android termux - updated and added -f for fast download

John-Russel:
i edit the update script you can change this


ThamognyaKodi (commented):
The changes seem mostly good but the ones that change some of the documentation remove a little bit of information that might be useful but overall I found it to have more pros than cons. But I think some of these might be improved but that is just my opinion.

ura43:
Please just change the default aria2 options to your improved ones in another PR and on master,  since the other branches are obsolete

PR #390, opened 2022-01-25
added -f for fast download

John-Russel:
with -f
Screenshot_2022-01-25-15-28-41-98
without -f
Screenshot_2022-01-25-15-29-09-21


ura43 (changes requested):
Nope. Make it the default. Who wants to download slow if fast downloads exist?
I also think -x 8 and -s 8 are more sane

PR #391, opened 2022-01-25
fixed android termux installation (mpv not included)

John-Russel:
~~also I uncomment the fast download to use in default~~
added -s for speedup download on low bandwidth and low cpu power

PR #393, opened 2022-01-25
Removing rm -rf's

RaynardGerraldo:


RaynardGerraldo:
> Looks good so far. Can you please also include the bugfix for Android from #391 ? The guy can't make a workable PR

yea im confused he just commited a "speedup download" , im just gonna try and implement what he did for the makefile

RaynardGerraldo:
fixed @ura43

Makefile
```diff
@@ -23,27 +23,28 @@ install:
 	chmod 0755 $(DESTDIR)$(PREFIX)/bin/ani-cli
 	echo "Installation successful (Mac OS)"
 uninstall:
-	rm -rf $(DESTDIR)$(PREFIX)/bin/ani-cli
+	rm $(DESTDIR)$(PREFIX)/bin/ani-cli
 	echo "Removal successful (Mac OS)"
 
 else ifeq ($(Platform), Android)
-TERMUX_BIN := /data/data/com.termux/files/usr/bin
```
ura43, line 30: This one makes me worry. We should probably assign PREFIX to that, what do you think?

ura43 (approved):

Makefile
John-Russel: PREFIX := /data/data/com.termux/files/usr/bin

ura43: Well current master doesn't use the variable at all, but that could just mean that that got broken along the way too

John-Russel: Termux already set PREFIX to /data/data/com.termux/files/usr
so adding PREFIX is useless i think

ura43: > Termux already set PREFIX to /data/data/com.termux/files/usr so adding PREFIX is useless i think

wdym Termux set it? As in a default Termux environment has that variable by default?

RaynardGerraldo: i tested it and it works with or without it

ura43: then we can leave it out

CoolnsX: > wdym Termux set it? As in a default Termux environment has that variable by default?

yup termux has that prefix by default.. and removing above line won't do any harm... I think this PR is good to go.

image

Issue #385, opened 2022-01-24
Menu in Loop has hardcoded colors


Derisis13:
It appears when playing the first/last episode of any anime. Tried to fix it, but couldn't do it in a POSIX-compliant way.

71zenith:
Can you explain it a bit? I can't reproduce the error.

ura43:
> Can you explain it a bit? I can't reproduce the error.

Check out his PR and run any episode 1 of any anime

pystardust:
@Derisis13 
What about not coding it even and odd, but use one function for menu, which tweaks a global variable on and off making it alternate

PR #405, opened 2022-01-25
fix: menu alternating was hardcoded

pystardust:


pystardust:
refer #385 for discussion

pystardust:
we need to tell people to use tabs over spaces (proper indents) and remove trailing white spaces

ura43:
> we need to tell people to use tabs over spaces (proper indents) and remove trailing white spaces

Yeah, lets search for a github action to lint that

ura43 (approved):

PR #408, opened 2022-01-26
Fix #389 and #422

### LINKED ISSUE #389: Non-interactive usage through flags
if you know the choice/anime_id and episode:
`ani-cli  boruto -a 3 -e 2`

searches for boruto looks for (english sub) and chooses episode 2 if exists then plays into mpv 

NO prompt asking to select episode!
Goes right to playing!

71zenith:
It cannot accept a range of episodes at this point and settings subs is not supported cause I cannot find any thing in the script about that.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -17,13 +17,14 @@ help_text () {
 	done <<-EOF
 
 	Usage:
-	  $0 [-v | -i] [-q <quality>] [-e <arguments>] [-d | -p <download_dir>] [<query>]
+	  $0 [-v | -i] [-q <quality>] [-e <arguments>] [-a <episode>] [-d | -p <download_dir>] [<query>]
 	  $0 [-v | -i] [-q <quality>] [-e <arguments>] -c
 	  $0 -h | -D | -U | -V
 
 	Options:
 	  -c continue watching anime from history
 	  -e pass arguments to the player/downloader
+	  -a specify episode to watch
 	  -h show helptext
 	  -d download episode
 	  -p download episode to specified directory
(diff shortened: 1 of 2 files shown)


71zenith:
Made so that if both episode and anime is passed, it will non-interactively select the first result and start playing. Only thing left is merging which breaks the former.

ani-cli
```diff
@@ -17,13 +17,14 @@ help_text () {
...
 	Usage:
-	  $0 [-v | -i] [-q <quality>] [-e <arguments>] [-d | -p <download_dir>] [<query>]
+	  $0 [-v | -i] [-q <quality>] [-e <arguments>] [-a <episode>] [-d | -p <download_dir>] [<query>]
 	  $0 [-v | -i] [-q <quality>] [-e <arguments>] -c
 	  $0 -h | -D | -U | -V
 
 	Options:
 	  -c continue watching anime from history
 	  -e pass arguments to the player/downloader
+	  -a specify episode to watch
```
ura43, line 27: - [ ] Add note that always first result is used, somewhere below the option list

ani-cli
```diff
@@ -251,32 +252,35 @@ menu_line_strong() {
...
-			selection_id=$anime_id
-			break
-		fi
-		count=$((count+1))
-	done <<-EOF
-	$search_results
-	EOF
+	if [ -n "$ep_choice_to_start" ] && [ -n "$select_first" ]; then
+		selection_id="$(echo $search_results | sed 's/\ .*//' )"
+	elif [ -z "$ep_choice_to_start" ] || { [ -n "$ep_choice_to_start" ] && [ -z "$select_first" ]; }; then
```
ura43, line 257: - [ ] excessive use of || and && inside an if else directive is bad form

ani-cli
```diff
@@ -307,11 +308,14 @@ episode_selection () {
...
-		is_half_ep=$(cut -c1-1 "$ep_choice_start")
-		if [ "$is_half_ep" = "h" ]
+		if [ -z "$ep_choice_to_start" ]; then
+			prompt "Choose episode" "[$first_ep_number-$last_ep_number]"
+			ep_choice_start=$REPLY
+			ep_choice_end=$REPLY2
+		else
+			ep_choice_start=$ep_choice_to_start && unset ep_choice_to_start
+		fi
+		if [ "$(echo "$ep_choice_start" | cut -c1-1)" = "h" ]
```
ura43, line 318: - [ ] Evaluate the cut command outside the if condition (SC2312)
- [ ] Unnescessary echo into cut. Just give the argument directly to cut

ani-cli
```diff
@@ -251,32 +252,35 @@ menu_line_strong() {
...
-		if [ "$count" -eq "$choice" ]; then
-			selection_id=$anime_id
-			break
-		fi
-		count=$((count+1))
-	done <<-EOF
-	$search_results
-	EOF
+	if [ -n "$ep_choice_to_start" ] && [ -n "$select_first" ]; then
+		selection_id="$(echo $search_results | sed 's/\ .*//' )"
```
ura43, line 256: - [ ] Don't echo pipe into sed
- [ ] Is it really impossible to put $search_results in quotes?

ura43 (changes requested):
- [ ] bump the version to 1.4.2

ani-cli
71zenith: how do i put it in good form?

ani-cli
71zenith: Cannot do either cause it would break.

ani-cli
71zenith: Did it

ani-cli
ura43: Eh its fine, just realized stuff like this is still common in the script. Thats a task for someone else during refactoring

ura43 (changes requested):
Tested working successfully.
Please skip the menu when running non-interactively, leading to the script just launching mpv and then exiting again.
That way its truly non-interactive

71zenith:
Did it

ani-cli
71zenith: Fixed it by using a different method

ura43 (approved):
Confirmed functional with `fooly-cooly ep 3`
I'm glad you got it shellcheck compliant

PR #403, opened 2022-01-25
Fix bug #388

71zenith:
Added -e flag to pass arguments to the player. Works with mpv. Updated the Readme

PR #410, opened 2022-01-26
Fix makefile to work with latest update #403

RaynardGerraldo:


RaynardGerraldo:
@ura43 can i just merge this? its a pretty simple fix

PR #415, opened 2022-01-26
Auto play

BAaboe:


RaynardGerraldo:
@BAaboe shellcheck is complaining, can you fix it so it passes the check?

`Notice: ./ani-cli:321:10: note: Double quote to prevent globbing and word splitting. [SC2086]
Notice: ./ani-cli:410:10: note: Prefer double quoting even when variables don't contain special characters. [SC2248]
Notice: ./ani-cli:419:10: note: Prefer double quoting even when variables don't contain special characters. [SC2248]`

BAaboe:
Checks if you have two nummbers in the episdoe input, if so it will enable auto play. When auto play is enabled it will make the video player pause the terminal until the video player closees aka when the episode is done, then it will wait two seconds so the user has time to quit the program in between episodes. it loops through all the episodes you specified and when it's done it goes back to normal ani-cli where it will ask the user if he wants to watch more or quit and so on

BAaboe:
@ura43  everything is resolved now

PR #420, opened 2022-01-26
New installation instructions

ura43:
To make the makefile obsolete

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,5 +104,5 @@ make
 
 ## Disclaimer
 
-The disclaimer of this project can be found [here](./disclaimer.md)
+The disclaimer of this project can be found [here.](./disclaimer.md)
 


sgtxd:
The new instructions should include the availability of the AUR package on Arch-based Linux distros

ThamognyaKodi (changes requested):
I suggest maybe moving the period outside the link like this `The disclaimer of this project can be found [here](./disclaimer.md).` since the period is not in the link. But moreover this is a much more minute change that does not add much value to the readablity.

ura43:
> I suggest maybe moving the period outside the link like this `The disclaimer of this project can be found [here](./disclaimer.md).` since the period is not in the link. But moreover this is a much more minute change that does not add much value to the readablity.

Good catch, I just change that back and forth whenever I need a diff to open a draft PR

ura43:
Opening another PR soon

PR #425, opened 2022-01-27
fixed the makefile for mac

iamchokerman:


ura43:
This will break it for Android, since `uname -s` evaluates as Linux on both Android and GNU/Linux

PR #430, opened 2022-01-27
Small bug introduced with auto play

BAaboe:
There was a small bug when you the video player closed at last episode of the auto play range, where the "Currently playing" message got displayed twice.


ura43 (changes requested):
bump to 1.4.4

PR #445, opened 2022-01-28
Fix #424

### LINKED ISSUE #424: Seem to be unable to pass multiple args to mpv with '-e' opt.
Version: 1.4.1 (ran a git pull about an hour before posting)
OS: Arch Linux kernel 5.16.2
Shell: Bash
Anime: itai no wa iya nano de... (non-dub)
mpv version: 0.34.1 w/ MPRIS plugin v. 0.6
youtube-dl version: 2021.12.17

**Describe the bug**
Tried multiple option syntax to pass the three options below to mpv, however the player refuses to load when there's more than 1 argument. 

**Steps To Reproduce**
./ani-cli -e --brightness=-15 --gamma=-15 --contrast=15 (results in usage text)
./ani-cli -e "--brightness=-15 --gamma=-15 --contrast=15"  (no errors appear, but player does not load)
./ani-cli -e '--brightness=-15 --gamma=-15 --contrast=15'  (same as above)
./ani-cli -e --brightness=-15 -e --gamma=-15 -e --contrast=15 (only 1 arg is passed)

**Expected behavior**
Pass multiple args to mpv

**Additional context**
Not a huge thing really as I can just put it in a config or modify $player_arguments directly in ani-cli but it would be nice to have. Thanks, awesome program

71zenith:


ani-cli
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # version number
-VERSION="1.5.4"
+VERSION="1.6.0"
```
ura43, line 4: Why all the way to 1.6.0 ?

ura43 (changes requested):
We still need to clarify how to correctly use -e in the helptext and Readme

ani-cli
71zenith: cause that was the milestone set in the issue, if you want I will lower it down

ura43: Yeah, please do. But honestly we have way to few issues right now

71zenith: Done, you can review

ura43 (approved):

PR #450, opened 2022-01-29
Multiple -e

71zenith:


ura43 (changes requested):
The point of multiple -e was no longer requiring xargs, does that make sense?

71zenith:
> The point of multiple -e was no longer requiring xargs, does that make sense?

Ohk, I'm still unclear how to get rid of xargs, but will give some try.

PR #452, opened 2022-01-29
update: Adding xargs as dependency

freezboltz:


71zenith:
please add xargs to the dependency checking

ura43 (changes requested):
As zenith said. If we list it as a dependency, we should also check for it in the script.

ura43:
Closing because -e is getting removed

PR #462, opened 2022-01-31
fix: Adding exit option in search anime menu

freezboltz:
organize code into function and adding exit and search another anime option into search anime menu.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -564,12 +564,13 @@ case $scrape in
 		search_results=$(search_anime "$query")
 		[ -z "$search_results" ] && die "No search results found"
 		anime_selection "$search_results"
-        [ "$REPLY" != "[0-9]+$" ] && episode_selection
         while [ "$REPLY" = "a" ]; do
             search_another_anime
         done
         [ "$REPLY" = "q" ] && exit 0
-        die "invalid choice"
+        if [ "$REPLY" -gt 0 ]; then episode_selection
+        else die "invalid choice"
+        fi
 		;;
 	history)
 		search_results=$(sed -n -E 's/\t[0-9]*//p' "$logfile")
--- a/ani-cli
+++ b/ani-cli
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # version number
-VERSION="1.6.1"
+VERSION="1.6.2"
 
 
 # history file path
--- a/ani-cli
+++ b/ani-cli
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # version number
-VERSION="1.6.0"
+VERSION="1.6.1"
 
 
 # history file path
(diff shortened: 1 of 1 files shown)


71zenith:
Doesn't work. Try fixing it

freezboltz:
> Doesn't work. Try fixing it

On it

freezboltz:
> Doesn't work. Try fixing it

fixed!

ani-cli
```diff
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # version number
-VERSION="1.6.0"
+VERSION="1.7.0"
```
ura43, line 4: You should bump to 1.6.1
I can understand why that may be confusing tho

ani-cli
```diff
@@ -544,7 +564,15 @@ case $scrape in
...
-		episode_selection
+        if [ "$REPLY" != "^[0-9]+$" ] ; then
+            while [ "$REPLY" = "a" ]; do
+                search_another_anime
+            done
+            [ "$REPLY" = "q" ] && exit 0
+            [ "$REPLY" != "[aq]" ] && die "invalid choice"
+        else
+		    episode_selection
+        fi
```
ura43, line 575: This could be 3 lines of `[ "$REPLY" = pattern ] && action`
Something like this could work
```suggestion
        [ "$REPLY" != "[0-9]+$" ] && episode_selection
        [ "$REPLY" = "a" ] && search_another_anime
        [ "$REPLY" = "q" ] && exit 0
        die "invalid choice"
```

ani-cli
```diff
@@ -613,21 +641,10 @@ if [ -z "$select_first" ]; then
...
 			;;
 		a)
-			tput reset
-			prompt "Search Anime"
-			query="$REPLY $REPLY2"
-			search_results=$(search_anime "$query")
-			[ -z "$search_results" ] && die "No search results found"
-			anime_selection "$search_results"
-			episode_selection
+            search_another_anime
```
ura43, line 644: Making these functions elsewhere in the script is a welcome change, as it makes our interactive part more readable.
Good job here

ura43 (changes requested):

ani-cli
ura43: :+1:

ani-cli
freezboltz: second time search is not working

ani-cli
freezboltz: we should have a guideline for version bumping

freezboltz:
after exit `illegal number error` is popping up should we redirect it or write addition checks.
shell: dash

ura43:
Btw this is not a bug fix. This is a feature request.
I'll make a UML behaviour diagram soon-ish to review our UX and will not merge this PR before that

ani-cli
ura43: I'm not sure how we should further clarify the version bump guideline.
If you have a good idea to expand 2. to be more actionable, please open another PR

ura43 (approved):

PR #465, opened 2022-02-01
refactor: use sed -n 4p instead of head | tail

Euro20179:


freezboltz:
this may fail
1 360p
2  480p
3 720p
if `sed` is implement with '4p' older anime may fail to parse

Euro20179:
Yeah, you right, current implementation is better then.

PR #466, opened 2022-02-01
Update README for Windows installation

RaynardGerraldo:
Scoop update

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,19 +119,21 @@ chmod +x $PREFIX/bin/mpv
 ```
 
 ### Windows
-
-* Download and install [dependencies](#Dependencies)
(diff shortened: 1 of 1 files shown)


ura43 (approved):
We may want to discourage -u usage

RaynardGerraldo:
What about the updates in between releases? e.g .1.6.1

ura43:
> What about the updates in between releases? e.g .1.6.1

As long as you can update with scopp after updating through -U, it's fine for me

RaynardGerraldo:
> > What about the updates in between releases? e.g .1.6.1
> 
> As long as you can update with scopp after updating through -U, it's fine for me

Of course,once ani-cli is installed using scoop, using the -U flag will update that file

Scoop update will run once we upload a new release e.g 1.7

PR #471, opened 2022-02-01
added so you can select all episodes when you select a range

BAaboe:
if you specify the range: 1 -1 it will automatically set the last ep in the range to the last ep in the anime


freezboltz:
Suggestion:
anime like one piece, we have 1000+ ep
add some termination mechanism for script
otherwise it will be annoying

freezboltz (commented):
for shell check use dash shell

ura43 (changes requested):
- [ ] Bump to version 1.6.1
- [ ] Document the feature in the help string
- [ ] Document the feature in the readme

BAaboe:
@ura43 it's fixed

ura43 (approved):

PR #480, opened 2022-02-02
fix: removing unnecessary newlines and fix proper indentation

freezboltz:


Derisis13:
Looking good, but while you're at it, could you please fix the indentations at lines 568-574 and 582-585

freezboltz:
> Looking good, but while you're at it, could you please fix the indentations at lines 568-574 and 582-585

on it

ura43:
I'd prefer rays PR for this

PR #487, opened 2022-02-03
fix: added proper indentation and removed newlines.

freezboltz:


Derisis13 (changes requested):
Changes are looking good to me, a few more white-space problems I would like to see solved:
lines 50, 189, 396: double spaces mid-line
lines 649, 652: spaces instead of tabs for indentation

That's all I found with a quick `ctrl+f`

ura43:
This seems to make a lot more sense as part of #489 
Closing

PR #486, opened 2022-02-03
docs: added tabwidth specification for consistency

freezboltz:

PR #489, opened 2022-02-04
1st half of february cleanup

Derisis13:
Part of my regular QA commits. I attempt to improve code readability and maybe consistency and code size.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -14,8 +14,8 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
-# 
-# Project repository: https://github.com/pystardust/ani-cli 
+#
+# Project repository: https://github.com/pystardust/ani-cli
 # Version number:
 VERSION='1.6.5'
 
@@ -354,14 +354,14 @@ check_input() {
 	# checks if input is number, creates $episodes from $ep_choice_start and $ep_choice_end
 	if [ "$ep_choice_start" -eq "$ep_choice_start" ] 2>/dev/null; then
 		true
-	else 
+	else
 		die 'Invalid number entered'
 	fi
 	episodes="$ep_choice_start"
 	if [ -n "$ep_choice_end" ]; then
 		if [ "$ep_choice_start" -eq "$ep_choice_start" ] 2>/dev/null; then
 			true
-		else 
+		else
 			die 'Invalid number entered'
 		fi
 		# create list of episodes to download/watch
@@ -569,9 +569,9 @@ case "$scrape" in
 		if [ "$REPLY" = 'q' ]; then
 			exit 0
 		fi
-		if [ "$REPLY" -gt 0 ]; then 
+		if [ "$REPLY" -gt 0 ]; then
 			episode_selection
-		else 
+		else
 			die 'invalid choice'
 		fi
 		;;
--- a/ani-cli
+++ b/ani-cli
@@ -112,16 +112,15 @@ search_eps () {
 search_for_unwatched () {
 	# compares history with gogoanime, only shows unfinished anime
 	search_results=$*
-
 	unwatched_anime=''
+	
 	while read -r anime_id; do
 		current_ep_number="$(search_eps "$anime_id")"
 		history_ep_number="$(sed -n -E "s/${anime_id}\t//p" "$logfile")"
 		if [ "$current_ep_number" -ge "$history_ep_number" ]
 		then
-			# breaks down with \n because of the display functions
-			unwatched_anime="$unwatched_anime$anime_id
-"
+			unwatched_anime='1'
+			echo "$anime_id"
 		fi
 	done <<-EOF
 	$search_results
@@ -129,7 +128,6 @@ search_for_unwatched () {
 	if [ -z "$unwatched_anime" ]; then
 		die 'No unwatched episodes'
 	fi
-	printf '%s' "$unwatched_anime"
 }
 
 search_another_anime () {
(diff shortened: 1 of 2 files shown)


Derisis13:
Should I update `CONTRIBUTING.md` to include a more detailed formatting guideline or should I open another PR?

71zenith:
> Should I update `CONTRIBUTING.md` to include a more detailed formatting guideline or should I open another PR?

I would recommend making another pr or just adding to #486

Derisis13:
So what it doesn't display here because I made changes locally before pulling iamchokerman's commits is that there was a commit to remove the literal newline in `search_for_unwatched`

ani-cli
```diff
@@ -1,25 +1,21 @@
...
+VERSION='1.6.5'
 
 
-# history file path
-logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
-auto_play=0
-
 #######################
 # Auxiliary functions #
 #######################
```
ura43, line 9: Can we replace such headings with `### TITLE ###` or `### Title ###` instead?

Derisis13: Of course!

Derisis13: Done in 95ebfa2

ani-cli
```diff
@@ -1,25 +1,37 @@
...
-
 #######################
-# Auxiliary functions #
+# AUXILIARY FUNCTIONS #
 #######################
 
 help_text () {
 	while IFS= read -r line; do
-		printf "%s\n" "$line"
+		echo "$line"
```
pystardust, line 29: why are you using `echo` here? 
Read this https://mywiki.wooledge.org/BashPitfalls#echo_.24foo
Avoid using echo cause the posix implementation of it is questionable. many shells have different implementations of echo even tho they are posix complaint.

ani-cli
```diff
@@ -117,103 +125,85 @@ search_eps () {
...
-	episode_selection
+	process_search
 }
 
 search_history () {
-	tput clear
-	search_results=$(sed -n -E 's/\t[0-9]*//p' "$logfile")
-	[ -z "$search_results" ] && die "History is empty"
-	search_results=$(search_for_unwatched "$search_results")
+	search_results="$(sed -n -E 's/\t[0-9]*//p' "$logfile")"
```
pystardust, line 150: why are you adding quotes here?
Shell word splitting is not applicable here. Word splitting does not occur in variable assignment.
source: posix sh specs 2.9.1 point 2

ani-cli
```diff
@@ -465,44 +435,44 @@ open_episode () {
...
-# to clear the colors when exited using SIGINT
-trap 'printf "\033[0m"; exit 1' INT HUP
+# default options
+player_fn='mpv' #video player needs to be able to play urls
+is_download="0"
+half_ep="0"
+quality='best'
+scrape='query'
+download_dir='.'
+choice=''
```
pystardust, line 469: I see that you have done a lot of
```
variable=''
```
The cleaner way would be
```
variable=
```
just leave a blank

ani-cli
```diff
@@ -465,44 +435,44 @@ open_episode () {
...
-
-# default options
-player_fn="mpv" #video player needs to be able to play urls
-is_download=0
-half_ep=0
-quality=best
-scrape=query
-download_dir="."
-choice=""
+touch "$logfile"
```
pystardust, line 477: why use touch?
It is an external utility. The script was using the beautiful inbuilts of shell scripting.

ani-cli
```diff
@@ -511,27 +481,24 @@ while getopts 'viq:dp:chDUVa:' OPT; do
 			exit 0
 			;;
 		d)
-			is_download=1
+			is_download='1'
```
pystardust, line 486: don't know numbers, we will later use these stuff in `-eq` statements in test. Its better you leave it as is, makes it evident that the variable is a number

pystardust (changes requested):
Try reading the up shell syntax(posix specifications). Quoting everything is definitely safe and nothing is wrong in it but you shouldn't do it when you don't need to.
There is really no reason to quote when there is not field splitting or we certainly know that the variable has a number

pystardust:
I wouldn't suggest this merge, maybe we could discuss here are open a new PR

ani-cli
Derisis13: It was more of a why not then why for me, but I can remove it if you wish

ani-cli
pystardust: basically
```
echo "hi\nthere"
```
word differently in dash and bash

ani-cli
Derisis13: Yes, it does appear cleaner, I'll change it

ani-cli
Derisis13: Shellcheck said not a thing, so I kept it, not knowing it's an external tool. I'll revert it.

ani-cli
Derisis13: You are absolutely right, I'll revert them

ani-cli
pystardust: Check out https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/, this is the specifications of shell scripting. Shell is beautiful, it has a grammar and then syntax that dictates what happens when.

ani-cli
Derisis13: I thought it was more readable, I wasn't aware how skimpy it is.

Derisis13:
> I wouldn't suggest this merge, maybe we could discuss here are open a new PR

I have implemented your suggestions, but I don't understand why a new PR? Did I change too much?

Derisis13:
Since a formatting guideline was decided not to be included in this PR, I'll mark it as ready in order not to create a bigger mess

Derisis13:
Testing has revealed significant issues, going forward with this PR would be more trouble than what it's worth. Closing PR.

PR #490, opened 2022-02-04
Fix #482

### LINKED ISSUE #482: Episode selection doesn't check for under-indexing
Version: 1.6.4
OS: Pop!_OS 21.10
Shell: bash
Anime: not applicable

**Describe the bug**
When you input a number lower than the specified range into the episode selection prompt, it tries to open the episode, which fails.

**Steps To Reproduce**

1. Run `ani-cli`
2. Search for any anime (eg. re:zero)
3. Choose something bellow the minimum ep. number (eg. -512)

**Expected behaviour**
I should receive an episode out of range warning and a new input dialouge

image

**Additional context**

71zenith:


ura43 (changes requested):
Can confirm working. Now gives a `Invalid number chosen` error.
However this also fails on `-1`, which should select the last episode as per #471

ura43 (approved):

PR #493, opened 2022-02-04
.5 episdoes fix for next and previous episode

BAaboe:


ani-cli
```diff
@@ -360,11 +360,29 @@ episode_selection () {
...
+	
 	if [ -n "$ep_choice_end" ]; then
 		[ "$ep_choice_end" -eq "$ep_choice_end" ] 2>/dev/null || die "Invalid number entered"
 		# create list of episodes to download/watch
 		episodes=$(seq "$ep_choice_start" "$ep_choice_end")
+
+		temp_episodes=""
+		for ep in $episodes
+		do
+			if curl -s "https://www1.gogoanime.pe/$anime_id-episode-$ep-5" | grep 404 > /dev/null; then
```
freezboltz, line 375: Avoid using absolute url instead use `base_url`

ani-cli
```diff
@@ -360,11 +360,29 @@ episode_selection () {
...
 	[ "$ep_choice_start" -eq "$ep_choice_start" ] 2>/dev/null || die "Invalid number entered"
+	
 	episodes=$ep_choice_start
+	[ "$half_ep" -eq 1 ] && episodes="h$ep_choice_start"
+	
 	if [ -n "$ep_choice_end" ]; then
 		[ "$ep_choice_end" -eq "$ep_choice_end" ] 2>/dev/null || die "Invalid number entered"
 		# create list of episodes to download/watch
 		episodes=$(seq "$ep_choice_start" "$ep_choice_end")
+
```
freezboltz, line 371: don't put space unless necessary

ani-cli
```diff
@@ -632,8 +657,19 @@ if [ -z "$select_first" ]; then
 	choice="$REPLY"
 	case $choice in
 		n)
-			ep_choice_start=$((episode + 1))
-			ep_choice_end=""
+			if curl -s "https://www1.gogoanime.pe/$anime_id-episode-$episode-5" | grep 404 > /dev/null; then
```
freezboltz, line 660: use `base_url`

freezboltz (changes requested):

BAaboe:
@freezboltz all fixed

ani-cli
```diff
@@ -632,8 +656,19 @@ if [ -z "$select_first" ]; then
...
+			else
+				if [ "$half_ep" -eq 1 ]; then
+					half_ep=0
+					ep_choice_start=$((episode + 1))
+					ep_choice_end=""
+				else
+					half_ep=1
+					ep_choice_start=$episode
+				fi
+			fi
```
pystardust, line 671: This approach can be improved.
What you are doing:
- checking if half ep of `$episode` exists
- incrementing by one if doesn't exist
- if it doesn't exist
  - increment it by one if half ep and remove half ep
  - add half ep if not half ep and add half ep
 
The better approach would be:
- check if `half_ep` equals 1, if yes then just increase `$episode` by one and set `half_ep` to 0.
- the rest remains ignoring `half_ep -eq 1` if part

So in the later approach you don't check if half ep exists if we are already on half ep, we will be saving an extra request.

BAaboe: i think i fixed it now

RaynardGerraldo:
Its working,but shouldnt we also implement for previous episode?

BAaboe:
> 

ah yes thats true, should be a easy fix

BAaboe:
> Its working,but shouldnt we also implement for previous episode?

this is working now

RaynardGerraldo:
> > Its working,but shouldnt we also implement for previous episode?
> 
> this is working now
yep,working

RaynardGerraldo:
> > Its working,but shouldnt we also implement for previous episode?
> 
> this is working now

yep working

ura43:
Can we not?
This seems like it bloats the code in an unacceptable way, considering how rare .5 episodes are and that you can still play them if you're aware of them

ani-cli
```diff
@@ -382,6 +397,14 @@ open_selection() {
 	# opens selected episodes one-by-one
 	for ep in $episodes
 	do
+		whether_half="$(echo "$ep" | cut -c1-1)"
+		if [ "$whether_half" = "h" ]
```
pystardust, line 401: if you want to compare the first character using shell then do this instead
```sh
 if [ ${str#h} != ${str} ]; then
    # it is half episode
    ... # do stuff here
else
    # it is not half ep
    ... # do stuff here
fi
 ```
basically if the string remains same after removing prefix h (if exists)

 
 > PS. Try not to use echo, it doesn't have concrete specifications, different shells act differently with it

pystardust:
also since the code is being repeated it would be better to have a next episode and previous episode functions that take in the current episode and and return the next one. Lesser code and more readable

ani-cli
pystardust: reason for doing so in the if statement is to avoid subshells and external utils when something could be done in pure sh.

BAaboe:
i just closed it since we are not going to use .5 episodes anyway

PR #502, opened 2022-02-06
add resume playback

nxfanaccount:
very simple modification to add a resume playback feature which can be disabled with `-s`.


nxfanaccount:
ignore for now, need to either scrap this or rework the logfile to include something like an "unfinished" tag

Issue #504, opened 2022-02-06
Empty string on different sections causes "integer expression expected" crashes.


kenielf:
**Metadata (please complete the following information)**
Version: Version: 1.6.6 
OS: Arch Linux (5.16.4-zen1-1-zen kernel)
Shell: zsh (/usr/bin/zsh)
Anime: "sono bisque doll wa koi wo suru" (but any other should do the same)

**Describe the bug**
Sending empty strings to the input sections crashes the program.
I tried this on both the anime selection and episode selection, with multiple anime and it seems to be constant.

**Steps To Reproduce**

1. Run `ani-cli` normally
2. When choosing an anime or an episode, instead of selecting normally, just press return to send an empty string.
3. It should return something along the lines of `./ani-cli: line 336: [: : integer expression expected`

**Expected behavior**
It should just loop the input step like when you insert an invalid number, but instead, it crashes.

**Screenshots (if applicable)**
![screenshot--2022-02-06--Sun-16-53-35](https://user-images.githubusercontent.com/61657416/152699055-aa81fe83-1d1c-4e52-88f4-9d676a3e7a3f.png)

**Additional context**
This can probably be easily remedied by adding an if clause to account for empty strings alongside the previous invalid number checks.

PR #511, opened 2022-02-07
fix: empty reply (#504)

freezboltz:
fix for #504 
by default added to first choice


pystardust (approved):

71zenith:
The issue asked for a menu in loop

freezboltz:
> The issue asked for a menu in loop

should we add loop ?

PR #513, opened 2022-02-08
February cleanup v2

71zenith:


Derisis13:
If we are implementing a function for is_number (which ended up being an one-liner) then I think we should implement it in a non-exiting manner, rather a test condition that can be inserted in the check of a while loop. Otherwise it's not a reduction of the codebase (which I intended it to be) but a bloat.

71zenith:
Ok, got it.

CONTRIBUTING.md
```diff
@@ -9,14 +9,37 @@
...
+### Discarded ideas and features:
+1. Option to edit the history file
+2. Auto-default to dub
+3. Resume playback
+4. Different provider(s)
+5. Makefile for installation
+6. iina support (until iina+ is out)
+
+PRs and issues relating to these will be closed without further notice. 
+If you want to know why we decided against these features, check closed PRs and issues.
```
ura43, line 42: Create a red "status: wontfix" label and close after applying that label

ura43 (approved):

CONTRIBUTING.md
ura43: That would also be good practice to put on declined feature requests the first time they come up

PR #520, opened 2022-02-09
fix: #504 and other fixes

71zenith:


iamchokerman (approved):

ani-cli
```diff
@@ -289,9 +289,14 @@ anime_selection () {
 		menu_line_strong "exit" "q"
 		prompt "Enter choice"
 		choice="$REPLY $REPLY2"
+        while ! [ "$choice" -eq "$choice" ] 2>/dev/null || [ "$choice" -lt 1 ] || [ "$choice" -ge "$count" ] || [ "$choice" = " " ] ;
+        do
+            [ "$choice" = "q " ] && die
```
freezboltz: I thick choice is not an error it shouldn't die instead `exit 0` is more appropriate

Derisis13: Nice find! I also think that exiting with 0 is what the script should do.

PR #525, opened 2022-02-12
Remove enhancement template

ura43:


ura43:
People seem to consistently confuse their feature request with enhancements.
Seems like true enhancements are internal only.
Let's hope people don't abuse bugreports for their feature requests now

iamchokerman (approved):

Issue #526, opened 2022-02-12
android termux issues


Crisp-Casper:
I received the error /data/data/com.termux/files/usr/bin/ani-cli: 475: cannot create /data/data/com.termux/files/home/.cache/ani-hsts: Is a directory, this was fixed with a "rm -rf /data/data/com.termux/files/home/.cache/ani-hsts" and script will run however mpv will not (when running mpv i get the error bash: /data/data/com.termux/files/usr/bin/mpv: Permission denied and I don't have super user to fix it)

Crisp-Casper:
> > I received the error /data/data/com.termux/files/usr/bin/ani-cli: 475: cannot create /data/data/com.termux/files/home/.cache/ani-hsts: Is a directory, this was fixed with a "rm -rf /data/data/com.termux/files/home/.cache/ani-hsts" and script will run however mpv will not (when running mpv i get the error bash: /data/data/com.termux/files/usr/bin/mpv: Permission denied and I don't have super user to fix it)
> 
> line 475 issue will be fixed by #527, as for "bash: /data/data/com.termux/files/usr/bin/mpv: Permission denied" can you try `chmod +x /data/data/com.termux/files/usr/bin/mpv`

ther termux chooser opens however on selecting either mpv or vlc

Crisp-Casper:
it crashes the android system... that close was on accident

RaynardGerraldo:
> it crashes the android system... that close was on accident

it crashes after selecting the players?

Crisp-Casper:
> > it crashes the android system... that close was on accident
> 
> it crashes after selecting the player?

I belive i changed a prefix attempting to find a solution after running "rm -rf ani-cli" and re-running the install script it works flawlessly

Crisp-Casper:
thank you very much for your help! very excited to have this on my phone

Crisp-Casper:
is there a way to support this project? I ain't looking to be a part of the team just pay for a coffee lol

RaynardGerraldo:
no problem, you can support ani-cli by simply starring the repo👍

ura43:
> is there a way to support this project? I ain't looking to be a part of the team just pay for a coffee lol

Follow the maintainers, it boosts our ego :)

If you really want to throw money at us, gift discord nitro to a maintainer

PR #527, opened 2022-02-12
fix: ani-cli not able to create history file if history dir doesnt exist

RaynardGerraldo:
#526 , line 475 bug

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="1.7.2"
+VERSION="1.7.6"
 
 #######################
 # AUXILIARY FUNCTIONS #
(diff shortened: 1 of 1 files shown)


Derisis13 (approved):
We're getting a lot of support issues for this, and it's a simple fix.

71zenith (approved):

ani-cli
```diff
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="1.7.1"
+VERSION="1.7.2"
```
ura43, line 21: Bump to 1.7.6 to account for missed bumps

ura43 (approved):

ani-cli
RaynardGerraldo: done

PR #530, opened 2022-02-12
fix : prevent exiting episode range by using previous episode

Goldenbeasty:
Prevent user escaping episode range by
1. Run `ani-cli evangelion`
2. Choose `1`
3. Choose `1`
4. Choose `p` (previous episode)

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -390,8 +390,9 @@ open_episode () {
 			err "Episode out of range"
 		fi
 		prompt "Choose episode" "[$first_ep_number-$last_ep_number]"
+		is_number "$REPLY"
 		episode="$REPLY"
-		ep_choice_start=$episode
+		ep_choice_start="$episode"
 	done
 	#processing half episodes
 	if [ "$half_ep" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -382,15 +382,16 @@ open_episode () {
...
 	is_number "$ep_choice_start"
 		if [ "$last_ep_number" -eq 0 ]; then
 			die "Episodes not released yet!"
 		else
 			err "Episode out of range"
 		fi
 		prompt "Choose episode" "[$first_ep_number-$last_ep_number]"
-		episode="$REPLY $REPLY2"
+		episode="$REPLY"
+		ep_choice_start=$episode
```
ura43, line 394: This probably breaks range selection

ura43 (changes requested):

ani-cli
Goldenbeasty: If I understand correctly range selection is not possible with episode variable anyways correct me if I'm wrong

RaynardGerraldo:
im sorry but we dont need this, ani-cli doesnt even show `[p] previous episode` when playing the first episode

plus ,users are not going to use the previous episode option when they know they are playing the first episode

ani-cli
BAaboe: this breaks the range selection

ura43:
Sorry but this seems pointless

Issue #417, opened 2022-01-26
Change the install process


ura43:
**In what category does your enhancement fall? [choose: safety, speed, ux, qa, other]**
qa

**Status Quo**
We currently document seperate install processes for each of the different plattforms we support, but still chose to refactor the Makefile to the point where now it is just a very verbose way of copying or removing ani-cli from PATH.

**The Enhancement**
Instead we could simply modify the install documentation to tell people to copy ani-cli to the path themselves + the mpv special case for termux.
This would change that step from being a black box ("oh cool, they have a makefile that installs it for me") to be more transparent ("I see, thats how I can run it from anywhere") and also more adaptable since users with weird M1 issues can easily just copy the file to the different PATH location.

**Additional context**
In the past month, issues surrounding installation and/or the makefile have caused a conservative estimate ~25% of the total issues, 90% of support inquiries, and at least 20 hours of work distributed amongst the maintainers.
It's easy to let yourself be dragged down by sunken cost.
If anyone can convince me how a `sudo make` oneliner is superior from a user perspective than a `sudo cp ani-cli /usr/local/bin/ani-cli`, I may reconsider my stance

ura43:
Slight correction: The makefile also marks ani-cli as executable. Doesn't change my opinion tho

sgtxd:
To also say something, the AUR package completely disregards the Makefile anyway and does the same thing as copying it into /bin and making it an executable.

Derisis13:
I personally never understood (why it exists) nor used the makefile... It also adds an extra dependency for the installation: `build-essential` (although it's pretty common to have it)

Xathros1:
if you want to work without the makefile, ok, but with the frequency of updates it is quite annoying to copy the file all the time.
Therefore I would suggest to work directly with a symlink.

So instructions for linux (other) would look like this:

```
git clone https://github.com/pystardust/ani-cli
cd ani-cli
chmod +x ani-cli
sudo ln -s /home/<username>/ani-cli/ani-cli /usr/local/bin
```

Uninstallation would then look like this:

`sudo unlink /usr/local/bin/ani-cli`

0x8FF:
we have already tried symlinks, so lol no @Xathros1

Xathros1:
As written, this was only a suggestion. In any case, it works.

sadkris:
On a tangent, Windows instructions need to be rewritten.

Git Bash doesn't allow users to use `$(WINDIR)` to reference environment variables.

[...]

PR #532, opened 2022-02-13
Add Makefile for linux and change Readme a bit

Zay-Ta:


71zenith:
We make already told that make file will not be brought back. It creates more problem that it solves. 
Keeping it open for sometime if you have some doubts

Zay-Ta:
yea i noticed that make file was removed from ani-cli but can u tell more about the problem it creates? I am just curious about it.

ura43:
It was about 20% of the issues and easily 50% of the support requests.
Fundamentally changing the install process from the black box that is the makefile to clear instructions has made it easier to help people with exotic setups.
We also made a very sophisticated one, with platform detection and stuff. But that didn't solve the problem either.

Lastly, I believe copy pasting sudo make is no easier than copy pasting a sudo cp

ura43:
For further context, read #417 or just search "makefile" in closed issues.

PR #534, opened 2022-02-14
docs: updating termux installation, now works with both mpv and vlc (-v option)

RaynardGerraldo:
using 'am start' again due to termux-open having different problems in different devices


iamchokerman:
tested and it worked

ura43 (approved):
We don't need the `chmod +x` iirc.
A bit more complex install process but better feature coverage.
Seems reasonable to me

RaynardGerraldo:
> We don't need the `chmod +x` iirc. A bit more complex install process but better feature coverage. Seems reasonable to me

For this installation method we do need chmod since we're using am start to instantly open mpv/vlc and not using termux-open to prompt a list of apps to choose from

Issue #542, opened 2022-02-17
Irregular Behaviour during Download


raj23689:
**Metadata**
Version: 1.8.0
OS: Axyl OS (Arch Linux)
Shell: fish
Anime: Fate Stay Night Heaven's Feel III Spring Song

**Describe the bug**
Downloading is broken.
It says something about an unsupported protocol, see screenshot.

**Steps To Reproduce**

1. Run `ani-cli -d fate night`
2. Choose 9
3. Choose episode 1

**Expected behaviour**
Downloading should work

**Screenshots**

![Screenshot_2022-02-17-09-10-04_1366x768](https://user-images.githubusercontent.com/92665411/154516986-9960a0ae-6530-412d-81db-13522b17c739.png)

PR #544, opened 2022-02-19
fix : link parsing problems and reduce curl call for majority of animes

### LINKED ISSUE #515: Unable to view an episode
Version: 1.6.7
OS: Windows 10
Shell: Git Bash 
Anime: enen-no-shouboutai-ni-no-shou

**Describe the bug**
Only this episode doesn't work the rest seem to work. Shows `Video URL not found`. Was able to replicate it on Termux.

**Steps To Reproduce**

1. Run `ani-cli`
2. Search fire force
3. Choose 4
4. Choose episode 1

**Expected behavior**
Video should play

image

RaynardGerraldo:
fix #515 #542

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ scoop install ani-cli -g
 Install termux [(Guide)](https://termux.com/)
 
 ```sh
-pkg install git make termux-tools ncurses-utils openssl-tool -y
+pkg install git termux-tools ncurses-utils openssl-tool -y
 git clone https://github.com/pystardust/ani-cli && cd ani-cli
 cp ani-cli $PREFIX/bin/ani-cli
 echo 'am start --user 0 -a android.intent.action.VIEW -d "$2" -n is.xyz.mpv/.MPVActivity' > $PREFIX/bin/mpv
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="1.8.0"
+VERSION="1.8.1"
 
 #######################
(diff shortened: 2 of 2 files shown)


71zenith (approved):

Derisis13 (changes requested):
Spaces on every single line, please change them to tabs. Also according to our current style guideline, `do` should be in the same row as `for`, just like `if` and `then`. I would also consider making  the `if ! ... then` into a `||`, if that makes sense for you. Otherwise looking good!

RaynardGerraldo:
> Spaces on every single line, please change them to tabs. Also according to our current style guideline, `do` should be in the same row as `for`, just like `if` and `then`. I would also consider making the `if ! ... then` into a `||`, if that makes sense for you. Otherwise looking good!

sorry bout that thanks for the review ,and thanks for the help @71zenith

CoolnsX:
@71zenith can u put ```-x 16 -s 16``` in line 446.. and remove commented lines above it...

71zenith:
> @71zenith can u put `-x 16 -s 16` in line 446.. and remove commented lines above it...

sure, but what shall be the commit message

71zenith:
tell before merging, we have to squash the commits

Derisis13 (approved):
Very nice! 
@Zenith71 you requested a reminder to squash

Derisis13:
Although line 159 should be part of line 158...

RaynardGerraldo:
> Although line 159 should be part of line 158...

@Derisis13 good?

Derisis13:
> > Although line 159 should be part of line 158...
> 
> @Derisis13 good?

Perfect

PR #551, opened 2022-02-24
Add old version warning for macOS

Candygoblen123:
Adds a warning about having an older, manually installed, version of ani-cli. Executables in `/usr/local/bin` can have precedence over those in `/opt/homebrew/bin`, meaning that users would continue using the older version, despite having the newer version installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,7 @@ Install homebrew [(Guide)](https://brew.sh/)
 brew tap iamchokerman/ani-cli
 brew install ani-cli
 ```
+*If you are upgrading from the old manual install process, you may have to remove the old ani-cli by running `rm /usr/local/bin/ani-cli`*
 
 ### Windows
 
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Install homebrew [(Guide)](https://brew.sh/)
 brew tap iamchokerman/ani-cli
 brew install ani-cli
 ```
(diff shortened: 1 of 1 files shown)


iamchokerman (changes requested):
it's a good idea to add that warning, however you need sudo permissions to remove something from outside your home directory so you need to add sudo in front of that command

Candygoblen123:
This would be true for most files in `/usr/local/bin/`, but because we copied the file over there we are still the owner of the file and have write permissions. I'll still add sudo anyway.

Here's some sample output from my terminal:
```sh
candy@MacBook ani-cli % cp ani-cli /usr/local/bin/
candy@MacBook ani-cli % sudo chmod +x /usr/local/bin/ani-cli
candy@MacBook ani-cli % stat /usr/local/bin/ani-cli
16777232 64632283 -rwxr-xr-x 1 candy admin 0 16278 "Feb 23 23:01:01 2022" "Feb 23 23:01:01 2022" "Feb 23 23:01:01 2022" "Feb 23 23:01:01 2022" 4096 32 0 /usr/local/bin/ani-cli
candy@MacBook ani-cli % ani-cli
Search Anime: beans
[1] pandalian-dub
[2] jack-to-mame-no-ki
[q] exit
Enter choice: q
candy@MacBook ani-cli % rm /usr/local/bin/ani-cli
candy@MacBook ani-cli %
```

iamchokerman (approved):

iamchokerman:
you are right, i actually didn't know that before, ty for letting me know with an example

PR #552, opened 2022-02-25
Wishlist: Add options to see history and check unwatched episodes only for the selected anime

newbee1905:


Derisis13 (changes requested):
That's a lot of loc for a feature with very little impact (as far as I'm concerned). Unless you can reduce the code and provide more insight on where it'd be useful, I can't consider merging.

Derisis13:
Also please use [semantic commits](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716) if possible, it greatly reduces our work when merging

ura43 (changes requested):
I'm also really sceptical about this.
What exactly are you trying to achieve?
And why does it matter?

newbee1905:
The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history.
Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking.
For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommend using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.

And sorry for not using sematic commits, I will add it next time after trying to reduce the code.

ura43:
> The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history. Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking. For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommend using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.
> 
> And sorry for not using sematic commits, I will add it next time after trying to reduce the code.

semantic commits while developing a feature aren't necessary. They are a burden we maintainers should carry when eventually merging something.
It sounds like you either have a very slow computer or rather high network latency.

I would see value in an alternative implementation of `-c` that just resumes the last anime you watched instead of making you select it from a list of all started, but not yet completed, anime.

Consider opening a feature request issue so that we can thoroughly discuss this before you spend more time writing code that probably won't be merged

newbee1905:
I still think the current -c is useful for people who don't watch anime every week to see what are the updates.

And thanks for the concern about me spending time on a feature that won't be merged. I am actually using this anyway, so even if it is not merged, I will probably keep it as a fork.

ura43:
> I still think the current -c is useful for people who don't watch anime every week to see what are the updates.
> 
> And thanks for the concern about me spending time on a feature that won't be merged. I am actually using this anyway, so even if it is not merged, I will probably keep it as a fork.

But having both would be cool, right?
Current `-c` definitely stays, but another way to continue is welcome

RaynardGerraldo:
> The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history. Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking. For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommendations using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.
> 
> And sorry for not using semantic commits, I will add it next time after trying to reduce the code.

So if im getting this right,you list out the animes in history, then only start searching until user selects it?

newbee1905:
> > The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history. Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking. For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommendations using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.
> > And sorry for not using semantic commits, I will add it next time after trying to reduce the code.
> 
> So if im getting this right,you list out the animes in history, then only start searching until user selects it?

yes and only for the one they select

RaynardGerraldo:
> > > The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history. Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking. For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommendations using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.
> > > And sorry for not using semantic commits, I will add it next time after trying to reduce the code.
> > 
> > 
> > So if im getting this right,you list out the animes in history, then only start searching until user selects it?
> 
> yes and only for the one they select

Then we could replace -c with this, its clearly faster and it works just like -c from a user perspective,right?

@ura43 @Derisis13

Derisis13:
> > > > The main purpose of this is to check for unwatched episodes for only one anime instead of the whole history. Using -c would run about 5s to 10s with a list of 5 animes in my history. So in the long run, the history would be too large and take a lot of time for checking. For example like, most people would remember which anime has a new episode on which day but the exact time is not consistent, so they would just want to check if that one anime has new episodes or not. I have chatted on discord and saw some recommendations using RSS but I don't think everyone would use RSS and going to gogoanime would be a hassle.
> > > > And sorry for not using semantic commits, I will add it next time after trying to reduce the code.
> > > 
> > > 
> > > So if im getting this right,you list out the animes in history, then only start searching until user selects it?
> > 
> > 
> > yes and only for the one they select
> 
> Then we could replace -c with this, its clearly faster and it works just like -c from a user perspective,right?
> 
> @ura43 @Derisis13

I like the idea of conserving bandwidth, but I don't quite understand how this would be any different from the deprecated `-H` function. Either we list all, or we need to scrape to decide what to display.

Also, since the specifications have changed a lot since the first idea, I'd suggest a new PR to have a fresh start.

RaynardGerraldo:
> I like the idea of conserving bandwidth, but I don't quite understand how this would be any different from the deprecated `-H` function. Either we list all, or we need to scrape to decide what to display.

ah,now i get it,my bad ignore what i said then

ura43:
I'll open a feature request for quick resume.
We won't revert to `-H`, but I want some way to just watch the next episode relative to the last thing I watched. Meaning no further input than `ani-cli -r` or whatever flag it will be.
I'll link that issue here once I wrote it

ura43:
#555

ura43:
Closing, continue discussion on #555

PR #554, opened 2022-02-25
Fix incompatibility with zsh

fold-squirrel:
My first commit ever yay

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -413,9 +413,7 @@ open_episode () {
 		sed -E "
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new"
-		if [ -n "$PID" ] && [ "$auto_play" -eq 0 ]; then
-			kill "$PID" >/dev/null 2>&1
-		fi
+		[ -n "$PID" ] && [ "$PID" -ne 0 ] && kill "$PID" >/dev/null 2>&1
 		[ -z "$video_url" ] && die "Video URL not found"
 		case "$player_fn" in
 			vlc)
--- a/ani-cli
+++ b/ani-cli
@@ -413,7 +413,7 @@ open_episode () {
 		sed -E "
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new"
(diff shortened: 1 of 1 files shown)


71zenith (changes requested):
Comply with shellcheck

ani-cli
```diff
@@ -413,7 +413,11 @@ open_episode () {
 		sed -E "
 			s/^${selection_id}\t[0-9]+/${selection_id}\t$((episode+1))/
 		" "$logfile" > "${logfile}.new"
-		kill "$PID" >/dev/null 2>&1
+
+        if [ "$PID" != "" ] && [ "$auto_play" -eq 0 ]; then
```
Derisis13, line 417: I would personally use `-z` instead of `!= ""`, much cleaner. Also, there's no reason for the empty lines, please remove them. The lines should be indented one further, since they are still part of the previous `if ` branch. This will be solved if you replace your spaces with tabs.

Derisis13 (changes requested):

Derisis13 (approved):
According to my testing, it works as expected. In `anime_selection` everything after the first space is not evaluated, but that's better than not being zsh compatible.

71zenith (changes requested):
revert back to the if statement, using one liners works but since we have multiple conditions, it'll cause ambiguity in future

ura43 (changes requested):
This doesn't break regular usage via dash.

Zsh works with the exception of episode ranges.
Is that fixable without unreasonable bloat?

ura43:
#555

ura43:
> zsh worked fine with episode ranges for me, I may have overlooked something in this commit, I'll try it and see where the problem is.

I tested with `flcl-dub` range 1-2

fold-squirrel:
I was working fine for me when I had sh linked to zsh and had `#!/bin/sh`, but when put `#!/bin/zsh` then it didn't work.
This is so confusing.

fold-squirrel:
okay I understand where the problem id and will work on it

ura43 (changes requested):
~~Please revert the last commit.
[-n PID] is preferable to initializing PID to 0.~~

Episode ranges still don't work.
If this isn't fixable / a lot harder to fix than what you have done so far, I'd prefer merging this without range support for the time being and opening a seperate issue for that.

ura43:
Had to bump your version due to merge order

ura43 (approved):

71zenith (approved):

Issue #523, opened 2022-02-11
Rejected Feature Requests


justchokingaround:
**1. Have a makefile** _#417_ 

**2. Resume playback from timestamp you were at last time** _#503_ 

**3. Change provider** _#383_ 

**4. Editing the hist file** _#499_ 

**5.  RSS feed integration** _#512_ 

**6.  Anilist / MAL sync** _(check out [trackma](https://github.com/z411/trackma))_

**7.  flatpak playersupport** _#484 (workaround available)_

**8.  Adding providers with a different language** _#593_

**9.  Extra options for mpv/vlc (eg. auto-fullscreen)** _#594_,_#598_

**10. Autocomplete** _#859_

**11. Flag for custom external downloader** _#854_

ura43:
8. RSS feed integration

Derisis13:
This should be reconsidered:
1: we did it anyway
6: --dub flag
9: done, default on steamdeck
10: idk if allanime has it, but the option is there if yes
11: possible with custom player function
13: same as 11

justchokingaround:
yep, i updated the text accordingly

Derisis13:
rip longest open issue

justchokingaround:
#i-was-there

justchokingaround:
auto complete sounds kinda interesting, shell completions are not hard to implement, i have a few ideas and resources

port19x:
> rip longest open issue

I figured pinned and closed is a thing, so that's neat.
That way we can get to zero open issues :smile:

justchokingaround:
clean

CoolnsX:
> 7 is already implemented. pls remove it

do you know, you can directly update other's messages?

PR #574, opened 2022-03-08
Add a Makefile again

PhantomDerp:
The Makefile was nice b/c I could update this without having to use a package manager, I can just run git pull and make whenever gogoanime decides to screw with it again.

this one just has 2 lines, one to copy to /usr/bin and one to git pull and then copy to /usr/bin

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,2 @@
+arch:
+	sudo cp ani-cli /usr/bin/ani-cli
--- a/Makefile
+++ b/Makefile
@@ -1,2 +1,6 @@
 linux:
 	sudo cp ani-cli /usr/local/bin/ani-cli
+
+linuxupdate:
+	git pull
+	sudo cp ani-cli /usr/local/bin/ani-cli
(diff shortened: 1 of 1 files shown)


iamchokerman:
check #523 before making PRs
and also, EVEN if we would've been adding a Makefile, u would have to make it mac and windows compatible. you're lucky i'm the one commenting, other maintainers, for other projects could've been very rude to u bc of this

ura43:
:joy:

First of all `ani-cli -U` will update ani-cli for you.

If only you knew how much trouble the makefile has caused us over the past several months before its removal.
We get less support requests than before, and the ones we do get are much easier to help.
Sure, if you really like gnu make for some reason, I won't tell you not to use a makefile.
But this wouldn't even realistically benefit 1% of our userbase.

Also this one is clearly mentioned in our [Rejected Feature Requests](https://github.com/pystardust/ani-cli/issues/523)
There are cool things to work on if you want to contribute, and I'll gladly point you in the right direction to do so, but this one will definitely not get merged.
I hope I sounded rude enough.

WONTFIX

ura43:
> check #523 before making PRs and also, EVEN if we would've been adding a Makefile, u would have to make it mac and windows compatible. you're lucky i'm the one commenting, other maintainers, for other projects could've been very rude to u bc of this

lol could

PR #586, opened 2022-03-16
Fix: Fix comparison with null returned ``PID``

Justaus3r:
While playing episodes following error is printed to stderr:
Screenshot_31
 
dunno the exact reason but maybe cuz the background job finishes or smth but shell var ``$!`` should at least be able to record the PID of the recently executed job.instead null is returned.well whatever.btw i reccom using string comparison instead of int for all shell vars cuz we don't always get wat we want cuz itzzz **Shell scripting**.

if anyone has a fix to null returned PID,please do edit the pr.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="1.9.2"
+VERSION="2.0.1"
 
 
 #######################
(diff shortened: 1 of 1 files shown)


Derisis13:
> While playing episodes following error is printed to stderr: Screenshot_31
> 
> dunno the exact reason but maybe cuz the background job finishes or smth but shell var `$!` should at least be able to record the PID of the recently executed job.instead null is returned.well whatever.btw i reccom using string comparison instead of int for all shell vars cuz we don't always get wat we want cuz itzzz **Shell scripting**.
> 
> if anyone has a fix to null returned PID,please do edit the pr.

How can I reproducethe bug? I haven't seen this before... I want to do some investigation, accepting a fix because "it works when I tested" is not a solution.

Derisis13 (approved):
Now I'd actually like to see this merged

PR #589, opened 2022-03-18
Add search another anime option on history.

meteor314:
Add search another anime option on history.
From this 
Anime

to

Anime so user can search an anime directly form history option.


Derisis13:
This might not seem much code but the fuction `search_another_anime` calls a lot of other functions resulting in bad spaghetti code. If we litter our codebase with calls like this maintaining will become a hell and that we do not want.

PR #598, opened 2022-03-28
Add fullscreen and screen args

rakusan2:
Adds args:
- `-f`
  - Enables fullscreen
- `-s`
  - Sets the screen to open mpv on
  - Does not work for vlc as I am unable to find the correct argument to set
  - Reason: I had an issue where mpv was opening on the wrong monitor

Reflectors play section:
- Easy to set arguments for different players
- Set optional arguments without tripping shellcheck
- Remove repetitive code

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -45,6 +45,8 @@ help_text () {
 	  -q set video quality (best|worst|360|480|720|1080)
 	  -v use VLC as the media player
 	  -D delete history
+	  -f fullscreen
+	  -s screen (0-32 integer)
 	  -U fetch update from github
 	  -V print version number and exit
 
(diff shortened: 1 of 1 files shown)


Derisis13 (commented):
We're holding all feature requests until v2 is merged, since it's our top priority.
Until then please tell us why this is preferable to doing this in your mpv config files or moving the window and double-clicking on it.

rakusan2:
Using the mouse for the same result gets tedious after doing it a few times, therefore I do not consider it as even an option.
On the other hand, using the config has the disadvantage of it being applied to all launches and not just when launched by ani-cli.

Benefits: 
- Both VLC and MPV have fullscreen arg therefore switching between them does not require making multiple configs
- Being able to set the screen enables the user to on the fly chose which monitor to have the player on without it being affected by which monitor the terminal is on

ura43:
Doesn't seem worth it over mpv config

ura43:
No

PR #601, opened 2022-03-29
Fix: New keys and IVs in GogoAnime

justfoolingaround:
_It is I_


ura43:
I'll bump the version and then we can merge

ura43 (approved):
Works for me

PR #609, opened 2022-03-31
Move play_episode section to a separate function

rakusan2:
There is no added/removed functionality

Purpose:
- Remove copypasta
- Increase readability

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -414,28 +414,7 @@ open_episode () {
 		" "$logfile" > "${logfile}.new"
 		[ "$PID" -ne 0 ] && kill "$PID" >/dev/null 2>&1
 		[ -z "$video_url" ] && die "Video URL not found"
-		case "$player_fn" in
-			vlc)
-				if [ "$auto_play" -eq 0 ]; then
-					nohup "$player_fn" --http-referrer="$dpage_link" "$video_url" > /dev/null 2>&1 &
-				else
-					inf "Currently playing$display_name episode" "$episode/$last_ep_number, Range: $ep_choice_start-$ep_choice_end"
-					"$player_fn" --play-and-exit --http-referrer="$dpage_link" "$video_url" > /dev/null 2>&1
-					sleep 2
-				fi
-				;;
-			*)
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -452,6 +431,30 @@ open_episode () {
 	fi
 }
 
+play_episode () {
+	# Build command
+	set -- "$player_fn" "$video_url"
```
Derisis13, line 436: I don't like the use of `set --` is there a reason you can't use a regular variable?

Derisis13 (changes requested):
v2 will be merged soon(tm), and we have decided to hold all other PRs until then. 
I can see merging this, but it'll need extensive testing.

ani-cli
rakusan2: I have looked into using arrays, but they are not supported under `/bin/sh` and regular variables do not vary the argument count.
On the other hand, `set --` uses the one array that all POSIX shells have. That being the arguments of the function it is called inside of.
That of course is also a downside as you need to keep in mind that each `set --` overwrites those arguments and `"$@"` contains them.

I was led to this solution by this wiki page https://github.com/koalaman/shellcheck/wiki/SC2086

rakusan2: After a bit more playing around, I had realized that it can be made a bit more readable by adding:
```sh
alias add_args='set -- "$@"'
alias set_args='set --'
```
That would enable the following changes
```sh
# from
set -- "$player_fn" "$video_url"
set -- "$@" --http-referrer="$dpage_link"
# to
set_args "$player_fn" "$video_url"
add_args --http-referrer="$dpage_link"
```
I had tested this and it works.
Do you want this in the PR?

Derisis13: I don't think you need an array. You can just use a simple string: `player_command="${player_fn} ${video_url}"` then add arguments like `player_command="${player_command} ${argumens}"`. Then execute it with "$player_command".

rakusan2: That passes a single argument that the player will not know what to do with.
If executed without the brackets, it expands to multiple arguments, but on every space regardless of quotes

I have tried if new lines change that, but they are treated the same as spaces

ura43:
I accidentaly closed this when deleting v2 after merging. please open another

PR #613, opened 2022-04-02
Move play-episode to separate function

rakusan2:
Same as #609 which was closed after v2 was merged


ura43:
Confirmed functional

ura43 (approved):
worthwhile refactor

Issue #610, opened 2022-03-31
"bad decrypt error, Video URL not found"


Gokurockx9:
Version: 1.9.3.
  OS: Windows 11 Build 20000.556
  Shell: git Bash
 Anime: Steins;Gate

Upon selecting any episode, it gives a bad decrypt error and it's unable to find the URL to the video. It was working a few days ago. 

Steps to reproduce is to select any episode, just as you normally would, to stream using mpv using Windows 11.

![Screenshot 2022-03-31 193248](https://user-images.githubusercontent.com/61856572/161166224-187c1465-f1cb-446e-a66a-aaac7511fd14.png)

vishal2376:
Getting same error in Ubuntu 20.04

HaibaoM55:
me too, but in windows 10

NotYasho:
same issue on Manjaro, tried reinstalling and deleting the config files, nothing changed. Any fixes?

NotYasho:
Solved it by updating to the latest version with the command
```console
sudo ani-cli -U
```

ura43:
Keep this one open for at least a few more days so we don't get more unnecessary duplicates

YuleTomalin:
Tried `scoop status` to see what the problem was. 
It says that **"aria2"** is outdated. Tried updating and reinstalling it, it's already in the latest version and it still didn't work. Any chance for a fix? Tried on both Powershell and Git Bash.

mexicancrisp:
based on earlier comment, `ani-cli -U` on Windows 10 in git bash worked for me as well

THEGOLDENPRO:
> Tried `scoop status` to see what the problem was. It says that **"aria2"** is outdated. Tried updating and reinstalling it, it's already in the latest version and it still didn't work. Any chance for a fix? Tried on both Powershell and Git Bash.
> 
> Edit: `ani-cli -U` on Windows 10/11 using Git Bash w/ Admin Privileges worked for me.

Thank you, that worked

ArekkusuMagni:
After adding sudo to ani-cli -U it was update & fixed for me.

HaibaoM55:
> After adding sudo to ani-cli -U it was updated & fixed for me. 

For windows is ./ani-cli -U

GetDeleted20454:
> Solved it by updating to the latest version with the command
> 
> ```
> sudo ani-cli -U
> ```

I love you <3

PR #617, opened 2022-04-03
Run patch with sudo if file is owned by root

rakusan2:
With this, the users won't need to run the entire script with `sudo` to update it

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -68,13 +68,22 @@ die () {
 	exit 1
 }
 
+patch_script () {
+	stat=$(ls -l "$0")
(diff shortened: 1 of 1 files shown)


71zenith:
Not very useful and we have doas users too

rakusan2:
Based on the conversation under #610 I think it is useful.
As for doas users, don't they still have sudo or a symlink with that name pointed to doas?

RaynardGerraldo:
Windows users wont benefit from this,since they need to open git bash with administrator to get sudo access

Also I'd rather run "sudo ani-cli -U" than adding 10 LOC to the script

PR #622, opened 2022-04-03
Android VLC

GlitchPhoenix98:
You can use the VLC android app for ani-cli using this command: echo 'termux-open --chooser --content-type=video "$2"' > $PREFIX/bin/mpv


CoolnsX:
We used to do that in the past.. but we ditched it for two reasons..
1. Doesn't work with every device..
2. Vlc does not support referrer links.. 

So it's a waste of time for u and me.. 
It will work for some anime.. but doesn't work for anime that gives vidstreaming links... We will consider it if gogoanime holds on the referrer thing..

GlitchPhoenix98:
Understood.

PR #626, opened 2022-04-05
reintroduce version metadata to issue template

ura43:
This appears to have been lost along the way to v2


ura43:
The fix attempts don't visually change anything and were unsuccessful

iamchokerman (approved):

Issue #636, opened 2022-04-09
What to do with piped input?


rakusan2:
Currently, piped input causes the script to break.

The following code can consume this input
```sh
if [ ! -t 0 ]; then
        # consume input
	pipe_input=$(cat)
        # close stdin
	0<&-
        # replace with tty
	exec 0</dev/tty
fi
```
The question after is what to do with it
1. Use it as a query
2. Use is as anime id
    - Possibly fall back to a querry
3. Discard it
4. Something else?

port19x:
good question

71zenith:
PIped input seems useless to me. ani-cli's main feature is that it's interactive.

rakusan2:
My intention is to simply add an input method that by my understanding is common while also removing a bug. 
It is not to take away any interactivity or functionality

For Example, with the PR #637 I can run:
- `dmenu -l 10 < ~./to_play_list | ani-cli` to easily connect dmenu to the script
   - equivelent to `chosen=$(dmenu -l 10 < ~./to_play_list); ani-cli "$chosen" `
- `curl 'http://anime.example/get/most/popular' | sed -n 4p | ani-cli` to have ani-cli query the 4th most popular anime
   - equivelent to `popular=$(curl 'http://anime.example/get/most/popular' | sed -n 4p); ani-cli "$popular" `
- `ani-cli -c "something"` in a script without worrying that a bug that redirects something will break the script or change its behavoiur

port19x:
I do see the value in that

port19x:
Have you clarify what bug this would fix?

rakusan2:
The bug is that if something gets piped in, the script prints `> Invalid choice entered` infinitely

rakusan2:
An idea as to how the piped input can be useful for you is in running test cases.
This is easy to do as each `read` command takes only a single line from what is piped in.

An environment variable can be used to enable such a mode and only the following changes need to be made to the code:
- any early stdin replacement with tty is disabled to allow a premade user input file to control the script
- mpv is set to exit early
- next video selection waits for the player to exit

port19x:
This would indeed be a very useful thing and could enable semi-automated testing

PR #637, opened 2022-04-09
Consume piped input

rakusan2:
This is a simple answer to my question #636 
This uses the first line of whatever is piped or redirected into the script as the search parameter

I have gone with this solution as it:
- adds only a couple of lines of code
- uses what is piped in
- solves a bug
- can be easily expanded on

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -451,6 +451,7 @@ scrape=query
 download_dir="."
 choice=
 auto_play=0
+pipe=
 # history file path
 logfile="${XDG_CACHE_HOME:-$HOME/.cache}/ani-hsts"
 logdir="${XDG_CACHE_HOME:-$HOME/.cache}"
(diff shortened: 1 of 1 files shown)


port19x:
this feature would bump ani-cli to `2.1.0`

port19x (changes requested):
- [ ] bump the version
- [ ] Don't use the `pipe` variable, directly store to query instead

port19x (approved):

Derisis13:
I'm very iffy on this one: this looks like one of those side-features that are going to be first sacrifised when we feel like we're getting bloated. I think an easier solution would be to just terminate if stdin was redirected.

rakusan2:
I find that a bit strange as piped input is fairly common on the tools I use.
Though if you do decide to sacrifice it, then it is just 4 lines of code that would be deleted and maybe a change of `elif` to `else` all of which is visible, separated, and easy to do.

Derisis13:
> I find that a bit strange as piped input is fairly common on the tools I use. 

Piping is useful, but sometimes just doesn't make sense. You don't pipe into VIM/nano (h)top, (t)ftp (and other file-sharing programs) or package managers for example. And there's xargs if you really want this feature...

port19x:
I fully agree that this is a less important feature at best.
The existance of xargs largely invalidates it, I didn't consider that factor.
If anything, I was hopeful that this may enable semi-automated testing through a seperate script, but that would then require many more lines to be a viable option.
After considering derisis input, I'm closing this one

PR #651, opened 2022-04-15
Set user-agent

rakusan2:
By default curl sets the header `user-agent` to `curl/<version>`
This means that the script announces itself as unwanted curl traffic

Setting this to a new value does not mean that the site will not figure out that this is not a browser

The `user-agent` in this PR is from my browser


port19x:
This is poorly implemented, and we're already bypassing many things anyway.
I see how you may think this helps, but sadly it really doesn't

PR #660, opened 2022-04-17
be able to select last episode only(without a range)

MMOOBB:
I noticed that whenever I try to select the last episode by passing in "-1" it wouldn't work and thought that was odd. Went through the code and figured it was because it only works if you pass in a range of episodes. I want to be able to quickly select the last episode and thought others may have thought the same. Just two simple lines added using an established method, no idea why there's that third line deleted and added.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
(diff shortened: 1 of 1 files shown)


port19x:
2 loc for the ability to select last ep with -1 seems worth it

port19x (approved):

PR #666, opened 2022-04-17
set fullscreen from commandline

MMOOBB:
Adds option to automatically fullscreen from the command line. Works for mpv and vlc.


iamchokerman:
no #523

PR #671, opened 2022-04-19
docs: Mac OS installation

iamchokerman:
Updated Mac OS installation instructions, I don't think a homebrew tap is necessary, and with all of the recent changes gogo  has made to their to their keys, it just makes more sense to install directly from source, instead of through latest releases

PS: please ignore the number of commits, i was doing some choking activity


port19x (approved):
with the recent behaviour of gogo I agree

Issue #665, opened 2022-04-17
Only audio playing


kalyan2012:
Version: 2.1.1
OS: Arch
Shell: zsh
Anime: any anime

**Describe the bug**
On playing the anime only sound plays in the background not the video
**Steps To Reproduce**
Just watch any anime

**Expected behavior**
Video and audio should be played

iamchokerman:
that sounds like an issue on your side, not with the script. what video player are you using with ani-cli: mpv or vlc?

kalyan2012:
In mpv only sound is played
In VLC just a new window opens no video no audio

iamchokerman:
what's the output of ````mpv --version``` and ```alias mpv```?

kalyan2012:
mpv --version
mpv 0.34.1-dirty Copyright © 2000-2021 mpv/MPlayer/mplayer2 projects
 built on UNKNOWN
FFmpeg library versions:
   libavutil       57.17.100
   libavcodec      59.18.100
   libavformat     59.16.100
   libswscale      6.4.100
   libavfilter     8.24.100
   libswresample   4.3.100
FFmpeg version: n5.0

alias mpv
no output

iamchokerman:
looks correct

iamchokerman:
try reinstalling ani-cli, also have you changed anything in the script itself?

kalyan2012:
ok ill try reinstalling, but will it effect my history?
I did not change

iamchokerman:
no it won't affect your history

kalyan2012:
For arch which might be good ani-cli or ani-cli-git?

iamchokerman:
ani-cli-git

kalyan2012:
Tried reinstalling but didnt work :(

kalyan2012:
Hey found some more details about the bug.
If i try to watch any anime from history the described problem is occurring
But if i start fresh it doesnt cause any problems. I guess i have to add them by rewatching into history

iamchokerman:
you could use the history transition script i suppose

Derisis13:
This could also be an m3u8 quality setting issue. Can you try playing the episodes that only had sound with `-q worst`?

kalyan2012:
> This could also be an m3u8 quality setting issue. Can you try playing the episodes that only had sound with `-q worst`?

Yup playing fine with worst quality. But why not with best?

MushroomEnder:
[...]

PR #674, opened 2022-04-19
feat: m3u8 quality with a different approach, #665 fix

RaynardGerraldo:


ani-cli
```diff
@@ -198,53 +198,48 @@ get_video_link() {
...
 			fi
 			;;
 	esac
 	printf '%s' "$video_url"
 }
 
+# select the $2th line in $1
+select_line() {
+	printf '%s' "$1" | head -n "$2" | tail -n 1
+}
```
Derisis13, line 219: Should this stay? This is something we tend to use more and more often, but it's a single line, so could very well be inline...

README.md
```diff
@@ -136,7 +136,7 @@ Install mpv-android [(Link)](https://play.google.com/store/apps/details?id=is.xy
 - curl
 - openssl
 - mpv - Video Player
-- aria2 - Download manager
+- aria2 - Download manager (neccessary for watching as well)
```
port19x, line 139: since when?

port19x (approved):

README.md
Derisis13: It stuck in from the first approach. I'll remove it.

Issue #679, opened 2022-04-20
cryptographic incompatibility with libressl


dvogiye:
**Metadata**
Version: latest from `master` branch; commit `48f2e20042945ee5298d92e6fab1cd1b21c414ee` (2.1.4)
OS: OpenBSD 7.1-beta
Shell: `/bin/sh` (`pdksh`)
Anime: Lain :)

**Describe the bug**
Guessing from the older reports and from the error message, it seems that the provider's keys have changed? 

**Steps To Reproduce**
For example, Lain :)
1. Run `ani-cli -d lain`
2. Choose 3 (Serial Experiments Lain)
3. Choose any #, i've tested with 5, 6, 7

**Expected behavior**
The script usually does work, even while using OpenBSD, with it's default LibreSSL and `pdksh` preferences. (btw, the script does in fact load with `pdksh`, I don't have `bash` installed on my system) I guess the answer to this is "it should work".

**Screenshots**
```
Loading episode 7... 
bad decrypt
9159946231440:error:06FFF06D:digital envelope routines:CRYPTO_internal:wrong final block length:/usr/src/lib/libcrypto/evp/evp_enc.c:540:
Downloading episode 7 ... 
: No such file or directory
Download failed episode: 007 , please retry or check your internet connection
(n) next
(p) previous
(r) replay
(s) select
(q) exit
> q
```

CoolnsX:
Working fine for me

muflona11:
> **Metadata**
> Version: latest from `master` branch; commit `48f2e20042945ee5298d92e6fab1cd1b21c414ee` (2.1.4)
> OS: OpenBSD 7.1-beta
> Shell: `/bin/sh` (`pdksh`)
> Anime: Lain :)
> 
> **Describe the bug**
> Guessing from the older reports and from the error message, it seems that the provider's keys have changed? 
> 
> **Steps To Reproduce**
> For example, Lain :)
> 1. Run `ani-cli -d lain`
> 2. Choose 3 (Serial Experiments Lain)
> 3. Choose any #, i've tested with 5, 6, 7
> 
> **Expected behavior**
> The script usually does work, even while using OpenBSD, with it's default LibreSSL and `pdksh` preferences. (btw, the script does in fact load with `pdksh`, I don't have `bash` installed on my system) I guess the answer to this is "it should work".
> 
> **Screenshots**
> ```
> Loading episode 7... 
> bad decrypt
[...]

PR #680, opened 2022-04-21
fix: token parsing

TheGlockMisc:
#679

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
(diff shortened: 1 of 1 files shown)


CoolnsX:
It's still working fine without that extra character u hv inserted.. it will always take all the string starting form that "token" word to the end...

TheGlockMisc:
> It's still working fine without that extra character u hv inserted.. it will always take all the string starting form that "token" word to the end..

Alright imma close this rn, thanks

CoolnsX:
No problem.. I recommend u should learn regex... It will be beneficial for u..

Issue #688, opened 2022-04-26
Can we mainline ani-sync?


port19x:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to watch anime with my friend again. Ani-sync was a great project, but has been archived about a month ago. I assume it may have been difficult and/or inconvenient to keep updated

**Describe the solution you'd like**
If this doesn't bloat the code in an unreasonable way, a `-s` for sync would be a great addition.

**Describe alternatives you've considered**
Consider resurrecting the third party repo and maintain it as a patch set over mainline ani-cli.

**Additional context**
syncplay would become an additional optional dependency.
I consider up to 100 loc reasonable for such a game changing feature.

iamchokerman:
that's actually a very interesting idea, @kalix123 aka Neptune and I were considering resurrecting ani-sync on his fork for ease of maintaining, but directly merging this into main as a new argument actually sounds much better. i'll look into it in the next few days, and implement it on a new branch, and will do some testing

YusufKhan-gamedev:
Please!

port19x:
@iamchokerman just pr the thing already :sweat_smile: 
Doesn't have to be range compatible on initial merge, we can patch that in shortly afterwards

iamchokerman:
ye sorry, i forgor 💀

PR #694, opened 2022-04-28
feat: ani-sync -s arg

iamchokerman:
#688

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -459,13 +459,8 @@ trackma_title="$(printf '%s' "$anime_id" | tr '-' ' ' | awk '{for(i=1;i<=NF;i++)
 			[ ! "$auto_play" -eq 0 ] && set -- "$@" "--play-and-exit"
 			set -- "$player_fn" "$video_url"  --http-referrer="$dpage_link"
 			;;
-		/Applications/Syncplay.app/Contents/MacOS/Syncplay)
-      if [ "$auto_play" -eq 0 ]; then
-        nohup $player_fn $video_url -- --referrer="$dpage_link" --force-media-title="$trackma_title $episode" > /dev/null 2>&1 &
-      else
-        inf "sharing $selection_id episode" "$episode/$last_ep_number, Range: $ep_choice_start-$ep_choice_end"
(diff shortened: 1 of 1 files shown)


port19x:
I'll give it some thorough testing  with a friend

port19x:
Seems to work, no issues so far

ani-cli
```diff
@@ -453,15 +453,17 @@ open_episode () {
...
-	set -- "$player_fn" "$video_url"
+trackma_title="$(printf '%s' "$anime_id" | tr '-' ' ' | awk '{for(i=1;i<=NF;i++){ $i=toupper(substr($i,1,1)) substr($i,2) }}1')"
 	case "$player_fn" in
 		vlc)
 			[ ! "$auto_play" -eq 0 ] && set -- "$@" "--play-and-exit"
-			set -- "$@" --http-referrer="$dpage_link"
+			set -- "$player_fn" "$video_url"  --http-referrer="$dpage_link"
+			;;
+		syncplay|/Applications/Syncplay.app/Contents/MacOS/syncplay)
+			nohup "$player_fn" "$video_url" -- --referrer="$dpage_link" --force-media-title="$trackma_title $episode" > /dev/null 2>&1 &
```
Derisis13, line 463: I'm actually a bit worried about this line's compatibility with auto-play mode (whenever you specify an episode range). It's fine to exclude that option from syncplay if it doesn't make much sense, what I'm worried about is the program ending up in an undefined state. I'd rather have it play from line 471 or 474

Derisis13 (changes requested):

ani-cli
iamchokerman: at the moment it's not compatible with auto-play mode, but ideally i think it should be imo. i'll look into this tomorrow, ty for the suggestion

iamchokerman: nvm, it was an easy fix, i just decided to use both my eyes and brain at the same time for once

iamchokerman: everything working as expected now, don't know why i complicated things in the beginning, sorry for that

Derisis13 (approved):
Now it passes

port19x (approved):
Confirmed working

Issue #693, opened 2022-04-28
Let's make a pull request template


port19x:
**Is your feature request related to a problem? Please describe.**
Currently our PR workflow feels kinda incomplete / awkward, a template could enrich it.

**Describe the solution you'd like**
Just have a checklist as detailed in the Contributing.md under Merge checklist

**Describe alternatives you've considered**
Adding the releases checklist seems like a good idea if we can communicate that it's not needed for every PR.
Additionally we could add a modified metadata section from the bug report template.

**Additional context**
I've had the idea for several weeks now.
A short ~google~ duckduckgo search gave me the following promising [article](https://embeddedartistry.com/blog/2017/08/04/a-github-pull-request-template-for-your-projects/
)

iamchokerman:
.github is probably a better location for "PULL_REQUEST_TEMPLATE.md" 🤔

port19x:
yes, definitely

PR #695, opened 2022-04-28
init(PULL_REQUEST_TEMPLATE.md)

iamchokerman:
this is just an early draft, don't merge this yet
#693


Derisis13 (commented):
I like the innitiative, but do we want the testing to be part of the PR template?

port19x:
> I like the innitiative, but do we want the testing to be part of the PR template?

Yes, to function as sort of a checklist

ani-cli
```diff
@@ -228,11 +228,11 @@ get_video_quality_m3u8() {
 		1080)
 			idx=5 ;;
 		best|*)
-			idx="$links_count" ;;
+			idx="auto" ;;
 	esac
 	if [ "$idx" -lt 1 ] || [ "$idx" -gt "$links_count" ];then
 		err "Current video quality is not available (defaulting to best quality)"		 
-		idx="$links_count"
+		idx="auto"
```
port19x, line 235: Is this the thing that fixes the mpv audio only?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+# Pull Request Template
+
+## Type of change
+
+Please delete options that are not relevant.
```
port19x, line 5: What is the reasoning behind this?

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
...
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected, or that adds new dependencies)
+- [ ] This change requires a documentation update
```
port19x, line 10: Just "documentation change" seems appropiate, we have those quite often

PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+# Pull Request Template
+
+## Type of change
+
+Please delete options that are not relevant.
+
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected, or that adds new dependencies)
```
port19x, line 9: Not sure if we need to elaborate in brackets there

PULL_REQUEST_TEMPLATE.md
port19x: Just wanted to comment one line here.
we should still have multiple things to tick.
Tho a checkbox that it's just a docs update would be a viable alternative

ani-cli
iamchokerman: ye

PULL_REQUEST_TEMPLATE.md
iamchokerman: agreed

PULL_REQUEST_TEMPLATE.md
iamchokerman: not that much reasoning, didn't pay attention when going over a template i copied

PULL_REQUEST_TEMPLATE.md
iamchokerman: yes

ani-cli
Derisis13: Wasn't this the only use of `links_count`? If yes, we should remove it's assignment as well

CoolnsX: We require that links count for if condition checking as if user gives out of range index then it will fall back to default, so...
However we can reduce it by merging both m3u8 links and links count variable

port19x:
I have moved testing from contributing.md into the PR template.
I'd also consider a checkbox for cryptographic key changes

port19x (approved):

PR #721, opened 2022-05-11
Allow episode selection before select

grey-C47:
# Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

# Description

This commit will allow episode selection before pressing "s" by inputting a number directly.
I believe the episode selection loop could be done without entirely / merged into the main loop, but I only added this much in this commit since I don't know if you guys agree with that.

# Checklist

- [x] any anime playing
- [x] bumped version

# Additional Testcases

- The safe bet: One Piece
- Not uploaded: one piece dub episode 900

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="2.2.0"
+VERSION="2.2.1"
 
 
 #######################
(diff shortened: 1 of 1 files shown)


port19x:
This makes no sense to me. Either you know what episode you wanna watch and use the flag for that, or you continue watching with -c or you're that 1% edge case we support because we can't justify not supporting a simple 2 loc feature

iamchokerman:
(loc = lines of code)

grey-C47:
@port19x I'm not sure I understand what you mean. To be clear this is to select an episode after already selecting a first one. When you are given the option to play prev, next or replay. Instead of pressing (s), and then writing the episode you want. You directly write the episode's number since it doesn't conflict with 'n', 'p', 'r' or 's' (and ideally remove 's').  
  
I thought this was clearly a better design, even if it is a small thing. I thought of this because I was going through episodes of an anime trying to find one I wanted to rewatch, jumping trhough a few episodes at a time.

port19x:
Yes, but this also circumvents explicit selection of single digit episode numbers unless I have misread.
Also explicit beats implicit.
I'd like to avoid this becoming that program where you misclick a key and get weird behaviour you didn't expect

PR #736, opened 2022-05-30
Fix tput and Illegal Number

rakusan2:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

- `tput` is a tool from ncurses which is not install by default on Android
- `$ep_choice_end` is empty when compared to a number
  - `"" <= 8`

## Checklist

- [x] any anime playing
- [x] bumped version
- [x] quality works
- [x] downloads work
- [x] syncplay -s works

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -18,7 +18,7 @@
 # Project repository: https://github.com/pystardust/ani-cli
 
 # Version number
-VERSION="2.2.0"
+VERSION="2.2.1"
 
 
 #######################
--- a/ani-cli
+++ b/ani-cli
@@ -68,6 +68,13 @@ die () {
 	exit 1
 }
 
+# tput is part of ncurses which is not installed by default on termux
+command -v tput >/dev/null || tput() {
+	if [ "$1" = 'clear' ]; then
+            printf '\033c'
+	fi
+}
+
 # get the newest version of this script from github and replace it
 update_script () {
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -398,6 +405,7 @@ episode_selection () {
 					err "Episode out of range"
 					continue
 				fi
+				[ "$ep_choice_end" = "" ] && break
```
port19x, line 408: I believe it is already checked during input wether ep_chice_end is empty

ani-cli
```diff
@@ -68,6 +68,13 @@ die () {
 	exit 1
 }
 
+# tput is part of ncurses which is not installed by default on termux
+command -v tput >/dev/null || tput() {
```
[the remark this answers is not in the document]
rakusan2: My intention is to only add `tput` if it does not exist
`>/dev/null` is there simply because `command` generates output that is not needed

ani-cli
rakusan2: The check likely did exist in a previous version, but in this version:
- `line: 387` sets it
- `line: 392` checks if it is `"-1"`
- `line: 401` converts it to a number and checks if it is less than `$ep_choice_start`

This change is between those lines and breaks out of the `while` before `line: 401` tries to convert `""` to a number

port19x:
The fix has been worked into v3 iirc
Either way, this PR is obsolete with the heavily refactored ani-cli

PR #741, opened 2022-06-06
full-fledged support for macOS with iina(better alternative to mpv for mac) && minor bugfix 

DaBigBlob:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [x] Documentation update

## Description

- iina is a much more feature-rich/less-hassle alternative for MacOS than mpv
- this pull req sets iina as default for macos using a case statement
- ensures compatibility with mpv flags and that all features of this script works with iina
### &&
```
if  [ "$ep_choice_end" -le "$ep_choice_start" ]; then
        err "Invalid range"
        continue
fi
```
👆bug fix for when $ep_choice_end is undefined [expected: number].

## Checklist

- [x] any anime playing
- [x] bumped version
- [x] quality works
- [x] downloads work
- [x] syncplay -s works


README.md
```diff
@@ -82,11 +82,26 @@ sudo chmod +x /usr/local/bin/ani-cli
...
 *Note that mpv installed through flatpak is not compatible*
 
+### Mac OS
+
+Install dependencies [(See below)](#Dependencies)
+
+Install [HomeBrew](https://docs.brew.sh/Installation) if not installed.
+
+```sh
+bash
```
justchokingaround, line 92: is bash really required here?

ani-cli
```diff
@@ -489,16 +489,20 @@ trackma_title="$(printf '%s' "$anime_id" | tr '-' ' ' | awk '{for(i=1;i<=NF;i++)
 		"syncplay"|"/Applications/Syncplay.app/Contents/MacOS/syncplay"|"/c/Program Files (x86)/Syncplay/Syncplay.exe")
 			set -- "$player_fn" "$video_url" -- --referrer="$refr" --force-media-title="$trackma_title $episode"
 			;;
+        iina)
```
justchokingaround, line 492: imo it would be preferable to check for the existence of iina in /Applications/ as well, since some people might already have iina installed, but not through homebrew; you can check how i used an or statement for syncplay to achieve that

ani-cli
```diff
@@ -513,7 +517,10 @@ trackma_title="$(printf '%s' "$anime_id" | tr '-' ' ' | awk '{for(i=1;i<=NF;i++)
 trap 'printf "\033[0m";[ -f "$logfile".new ] && rm "$logfile".new;exit 1' INT HUP
 
 # default options
-player_fn="mpv" #video player needs to be able to play urls
+case "$OSTYPE" in
+    darwin*) player_fn="iina";; #MacOS specific
```
justchokingaround, line 521: imo it would be preferable to check for the existence of iina in /Applications/ as well, since some people might already have iina installed, but not through homebrew; you can check how i used an or statement for syncplay to achieve that

justchokingaround (commented):
overall i think u did a great job, however we've already been reluctant to add iina as an additional player in the past (although it is great). additionally, i'm not sure it is a good idea to set iina as the default player for mac, since some people (myself included) have been using mpv and have a collection of mpv scripts they'd like to use. atm iina doesn't have good support for mpv user scripts. this among other reasons, is why i was reluctant to add iina support until iina plus comes out. also ani-cli v3 is in the works so this might not be the best timing to add a new video player. if u know something that i don't, please do tell 🙏

README.md
DaBigBlob: someone [like me] might use non-posix compliant shell like fish as their default; might result in unexpected/most likely behavior;

ani-cli
justchokingaround: yes ofc, homebrew is required for the other dependencies, however some people might have already installed iina previously, so to make it more user friendly it'd be better to have support for regular installations as well

DaBigBlob:
before i got a macbook pro with M1 pro a few months ago, i used gnu/void linux for 5 years and absolutely loved my time with mpv. but on macos arm i would like to use something native however the --cask version of mpv is x86 only and requires me to install rosetta. wouldnt want that overhead while im only watching a video. 🙁
could you at least add this👇 part?
```sh
iina) set -- "$player_fn" "$video_url" --no-stdin --keep-running --mpv-vid="$idx" --mpv-referrer="$refr" --mpv-force-media-title="$trackma_title $episode"
```

justchokingaround:
> 

i am actually in support of adding iina back, but probably in v3. could you please make a PR for the v3 branch? also i had no idea mpv was problematic on M1, now i know 🙏

PR #781, opened 2022-06-21
Added man page

CountOmega:


YusufKhan-gamedev:
Please dont add the compressed page directly here, instead its better to do [what grate does](https://github.com/grate-driver/xf86-video-opentegra/tree/master/man)  in its ddx when compiling man pages.

justchokingaround (approved):

PR #801, opened 2022-06-26
Add custom UI support and dmenu UI

rakusan2:
## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

With the UI section being in a separate file, it is easy to apply additional UI's over it.
This adds an option for that along with a custom UI to make use of it

This extra UI is for `dmenu`
With this: 
- `dmenu` handles the selection menu.
- it also takes the prompt input which is then split into `REPLY` and `REPLY2` same as what `read` does
- selection menu is shown only after the player closes
- input is still printed to console to not change the look of it

The new UI is used with `ani-cli -u dmenu`

## Checklist

- [x] any anime playing
- [x] bumped version
- [x] quality works
- [x] downloads work
- [x] syncplay -s works

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/bin/ani-cli
+++ b/bin/ani-cli
@@ -29,6 +29,7 @@ help_text () {
 	  -c continue watching anime from history
 	  -h show helptext
 	  -D delete history
+	  -u load UI
 	  -U fetch update from github
 	  -V print version number and exit
 	  -f select provider to scrape first [1-7]
@@ -482,7 +483,7 @@ logdir="${XDG_CACHE_HOME:-$HOME/.cache}"
 # source the UI
 . "$lib_dir/UI" || die "No UI file"
 
-while getopts 'svq:dp:chDUVa:xf:' OPT; do
+while getopts 'svq:dp:chDUVa:xf:u:' OPT; do
 	case $OPT in
 		d) . "$lib_dir/player_download" || die "No download player script found" ;;
 		a) ep_choice_to_start=$OPTARG ;;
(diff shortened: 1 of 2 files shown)


Derisis13:
@rakusan2 I created a wiki page for the UI script, hopefully it is helpful.

rakusan2:
My intention was to have this run over the existing UI and only change the prompts. That is why the input gets printed to the console after.
I did not want to copy anything that would not change.

Do prefer for this UI to be completely its own thing instead of reimplementing what differs?

port19x (changes requested):
No.
I don't think custom UI support is in any way justified.
However a dmenu UI could be useful to some
Maybe you can both support dmenu and rofi?

rakusan2:
How would you like for the `dmenu` UI to be turned on or switched with the default UI?
`dmenu` currently does not work under Wayland and therefore needs to be switchable if `UI_dmenu` is to be added

Solutions:
1. Use `-u` option to select the UI
2. Use the ENV variable `ANI_CLI_UI` to select the UI
3. Use a symlink
    - This would entail renaming `UI` to `UI_default` and if `UI` does not exist, linking it to `UI_default`
    - Update would not touch this symlink

Since solution 1 is off the table, what about 2 or 3?

Derisis13:
Since the script was re-merged into a single file, external module support is depecated.

PR #808, opened 2022-07-05
Update player_vlc

McUberStein:
Added redirect of all VLC output to /dev/null, because VLC standard output pushes the ani-cli menu out of the way, which is quite annoying.


justchokingaround (approved):
personally tested, makes complete sense

PR #812, opened 2022-07-08
feat: menu autoplay option

HotMonkeyWings:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

This allows users to select an autoplay option from the menu. It will keep autoplaying until the anime has reached the final episode. Just like next, it is hidden on the final episode.

## Checklist

- [x] any anime playing
- [x] bumped version
- [x] quality works
- [x] downloads work
- [ ] syncplay -s works

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/bin/ani-cli
+++ b/bin/ani-cli
@@ -2,7 +2,7 @@
 
 # License preamble at the end of the file
 # Version number
-VERSION="3.2.1"
+VERSION="3.2.3"
 
 
 #######################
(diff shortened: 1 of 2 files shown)


port19x (changes requested):
Problem is, we alrady have such a function: ranges
When you pass -1 as the second range parameter it plays the whole thing

HotMonkeyWings:
Yes, but that's a few extra steps right? This is in case somebody is interested in continuously bingeing instead of writing those inputs. Just figured it could save time

PR #816, opened 2022-07-11
Switched to axel as the anime downloader.

Si-Ki:
# Even faster downloading with AXEL

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

Supports multi-threading and is lighter and loads faster than, aria2.


Si-Ki:
I'm sure there is a way to feed all the links to `axel` at the same time as it supports that, which could give even faster speeds, commenting it cuz I can't do it and someone else must.

justchokingaround (approved):

PR #832, opened 2022-07-20
readme: better adhere to unix/arch file hierarchy

YusufKhan-gamedev:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ x ] Documentation update

## Description

The arch file hierarchy does not mention /usr/local at all(most likely because arch doesnt deal with local installs) and all of the distros I personally dealt with that do local installs do it in the home folder instead of /usr/local. If the user of ani-cli needs to do local installs they will most likely notice that we are doing installs in the /usr directory instead of wherever they place their binaries and will adjust the instructions to their needs.

The others forms were not applicable for a documentation update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,9 @@ For more info on providers, please refer to [this](https://github.com/pystardust
 Install dependencies [(See below)](#Dependencies)
 
 ```sh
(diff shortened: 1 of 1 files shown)


Derisis13:
No we don't want to deal with this. You install wherever you wish, but don't confuse the users with more changes regarding installation.

PR #867, opened 2022-08-15
fix: reduce axel connections 30 -> 10

### LINKED ISSUE #865: option to allow for max number of connections when downloading
When I normally download anime with ani-cli axel tries to download an episode it gets limited as the max connections are too high and it creates a bug where I cannot download anything and it simply repeats `50%` and `Connection 0 finished` forever. However, when I changed the number of connections in the axel download command to 10 and it worked perfectly. I've tried using axel to download other things with the same parameters and it works fine.

**TLDR;** _30 max connections is too much. And axel gets limited, causing it to not download anything. Changing it to 10 fixes it_

A flag that changes the number of connections made to the site when downloading. For example: `-n specify number of connections when downloading`

**Describe alternatives you've considered**
Manually update the download parameters as needed, and it works just fine. However, I wanted to create a feature request in case in the future anyone has similar issues.

**Additional context**
I think generally for most locations 30 max connections works just fine, but for the rare use cases like myself, it might be helpful to add on for anyone in the future to prevent this bug.

port19x:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

closes #865 

## Checklist

- [ ] downloads work

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -2,7 +2,7 @@
 
 # License preamble at the end of the file
 # Version number
-VERSION="3.3.2"
+VERSION="3.3.3"
 
 
 #######################
--- a/ani-cli
+++ b/ani-cli
@@ -140,7 +140,7 @@ download () {
 		*m3u8*)
 			ffmpeg -loglevel error -stats -referer "$1" -i "$2" -c copy "$download_dir/$3.mp4" ;;
 		*)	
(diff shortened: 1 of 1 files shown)


port19x:
how embarassing. forgot to bump the version

Derisis13 (approved):
Not much to say, it is what it needs to be.

PR #874, opened 2022-08-25
Scoopy

Derisis13:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Documentation update

## Description

Install and uninstall instructions for scoop

## Checklist

- [ ] any anime playing
- [ ] bumped version
- [ ] quality works
- [ ] downloads work
- [ ] syncplay -s works

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,12 +149,18 @@ brew install --cask iina
 
 ### Windows
 
-[todo]: <> (Scoop installation and notation of dependencies having to been installed trough scoop)
-
 *Make sure git bash is installed [(Install)](https://git-scm.com/download/win)*
 
 *Note that the installation instruction below must be done inside **Git Bash**, not in Command Prompt or Powershell*
 
+#### Scoop bucket
+
+```sh
+scoop bucket add extras
+scoop install ani-cli
+```
+
+#### From source
 ```sh
 rm -rf "/usr/local/share/ani-cli" "/usr/local/bin/ani-cli" "/usr/local/bin/UI" /usr/local/bin/player_* #If some of these aren't found, it's not a problem
 git clone "https://github.com/pystardust/ani-cli.git" && cd ./ani-cli
--- a/README.md
+++ b/README.md
@@ -204,7 +204,6 @@ sudo rm -f /etc/apt/trusted.gpg.d/ani-cli.asc /etc/apt/sources.list.d/ani-cli-de
 * .rpm
 ```
 sudo dnf remove ani-cli      # for ani-cli
-sudo dnf remove ani-cli-full # for ani-cli-full
 # disable the repo in dnf
 dnf copr disable derisis13/ani-cli
 ```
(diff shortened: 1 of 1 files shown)


Derisis13:
I want a confirmation on this that it works, I don't have a windows PC ready to test...

Wiener234:
> I want a confirmation on this that it works, I don't have a windows PC ready to test...

I can confirm that it works on windows 11 and would suggest that there be an info added to install mpv from scoop.

Derisis13:
> I can confirm that it works on windows 11 and would suggest that there be an info added to install mpv from scoop.

Does it only need to be explicitly installed when building from source (and the scoop package handles it as a dependency) or do yo explicitly need to install it in both cases?
Also the same is true for vlc and syncplay, right?

Wiener234:
> > I can confirm that it works on windows 11 and would suggest that there be an info added to install mpv from scoop.
> 
> Does it only need to be explicitly installed when building from source (and the scoop package handles it as a dependency) or do yo explicitly need to install it in both cases? Also the same is true for vlc and syncplay, right?

You need to install mpv, vlc and syncplay in every case serperatly. And when installing ani-cli via scoop mpv not installed via scoop doesnt work out of the box. I think i have read somewhere on discord that you can add mpv to path but i cant find that, same would probably go with vlc that also dosent work if not installed via scoop. 

For Syncplay installing it via scoop doesnt work, because it has to be in C:\Program Files(x86)\Syncplay\Syncplay.exe where ani-cli is looking for it but scoop doesnt install it there.

Derisis13:
Thank you, this is very useful information

README.md
```diff
@@ -105,11 +106,21 @@ sudo dnf install ani-cli
 
 #### Arch
 
+Build and install from the AUR: 
+```sh
+yay -S ani-cli
+```
 Also consider ani-cli-git
 
+#### OpenSuse Tumbleweed
```
port19x, line 115: I would assume this works in opensuse leap as well, whats up with that?

README.md
```diff
@@ -105,11 +106,21 @@ sudo dnf install ani-cli
...
+
+First you need to add the ani-cli copr repo, then install the package:
+
 ```sh
-yay -S ani-cli
+zypper addrepo https://download.copr.fedorainfracloud.org/results/derisis13/ani-cli/opensuse-tumbleweed-x86_64/
+zypper install ani-cli
 ```
+*Note: package is noarch, so any architecture should work, even though the repo is labled x86-64*
```
port19x, line 123: Is it possible to label the repo as noarch?
Can confirm ani-cli works on aarch64 btw

port19x (approved):

README.md
Derisis13: Absolutely no idea. I haven't even tested the package on tumbleweed, I just assume it works... Worst case I make a chroot for the different versions of leap too.

PR #879, opened 2022-09-02
change shebag and remove `shellcheck disable=SC2046`

MrRawes:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

the reason the shebang is changed from `/bin/sh` to `/usr/bin/env sh` is so that `sh` does not need to be in `/bin`, the reason of removing `shellcheck disable=SC2046` is that shellcheck does not complain about that

## Checklist

- [ ] any anime playing
- [ ] bumped version
- [ ] quality works
- [ ] downloads work
- [ ] syncplay -s works

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -13,4 +13,4 @@ jobs:
     - name: Run ShellCheck
       uses: ludeeus/action-shellcheck@master
       env:
-        SHELLCHECK_OPTS: -s sh -o all -e 2250 -e 2016 -x
+        SHELLCHECK_OPTS: -s sh -o all -e 2250
--- a/ani-cli
+++ b/ani-cli
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
 
 # License preamble at the end of the file
 # Version number
@@ -552,7 +552,6 @@ while getopts 'svq:dp:chDUVa:xr:fn' OPT; do
 done
 shift $((OPTIND - 1))
 progress "Checking dependencies.."
-# shellcheck disable=SC2046
 dep_ch "curl" "sed" "grep" "openssl" || true
 
 if [ "$player_fn" = "download" ];then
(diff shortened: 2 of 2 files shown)


Derisis13:
Why do you find it a good idea to change the shebang? Is `/usr/bin/env/sh` present on more systems than `/bin/sh`?

MrRawes:
> Why do you find it a good idea to change the shebang? Is `/usr/bin/env/sh` present on more systems than `/bin/sh`?

it is so the user could have `sh` in somwhere other then `/bin`

port19x:
swappy-20220904_205851

Your change would break defaults

port19x:
While verifying the shellcheck thing I noticed SC2016 is violated on line 155 and instead of using a shellcheck disable comment like you're supposed to we have that disabled in the workflow definition

.github/workflows/master.yml
```diff
@@ -13,4 +13,4 @@ jobs:
     - name: Run ShellCheck
       uses: ludeeus/action-shellcheck@master
       env:
-        SHELLCHECK_OPTS: -s sh -o all -e 2250 -e 2016 -x
```
port19x, line 16: note that we no longer need -x either due to our merge back to one file

ani-cli
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/usr/bin/env sh
```
port19x, line 1: This breaks the script on a normal arch installation

port19x (changes requested):

ani-cli
Derisis13: Similar problems might be present on termux

MrRawes:
> swappy-20220904_205851
> 
> Your change would break defaults

that screenshot does not have a space between `/usr/bin/env` and `sh`

port19x:
> > swappy-20220904_205851
> > Your change would break defaults
> 
> that screenshot does not have a space between `/usr/bin/env` and `sh`

good point, but I wouldn't call env to ask for sh when /bin/sh is standard and set everywhere by every programm that uses the shell

PR #912, opened 2022-10-11
fix: search gogoanime with updated url

JammSpread:
# Pull Request Template

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description
ani-cli currently displays the error "No search results found" when searching for an anime. It turns out the reason for this issue is that the domain name used, "https://gogoanime.lu" is no longer valid. Instead it tries to redirect you to a new domain, "https://gogoanime.dk". This new domain is now used in this bug fix.

## Checklist

- [X] any anime playing
- [ ] bumped version

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -2,7 +2,7 @@
 
 # License preamble at the end of the file
 # Version number
-VERSION="3.4.1"
+VERSION="3.4.2"
 
 
 #######################
--- a/ani-cli
+++ b/ani-cli
@@ -151,7 +151,7 @@ download () {
 # gets anime names along with its id for search term
 search_anime () {
 	search=$(printf '%s' "$1" | tr ' ' '-' )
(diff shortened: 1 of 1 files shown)


CoolnsX:
The -L argument in curl u edited is for redirection.. so if the link is redirecting to another it will follow the redirection

Derisis13:
If we were to merge this it would require a minor version bump. I'm not against it  though (that domain will expire once anyway).

JammSpread:
> The -L argument in curl u edited is for redirection.. so if the link is redirecting to another it will follow the redirection

Oh, that's nice. This PR was made as of yesterday when there was a faulty SSL certificate for the redirect site that prevented it from working as it does now.

Derisis13 (approved):
New URL is correct, version is bumped.

PR #937, opened 2022-12-11
No more dependencies!

metafates:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Documentation update
- [x] Improvement!

## Description

TLDR; no more dependencies!

**What?**

This PR specifies the dependencies for the new cross-platform tool called [tea](https://github.com/teaxyz/cli) (from the creator of [brew](https://brew.sh)).

> tea is not a package manager.
>
> tea is unified packaging infrastructure.
>
> From the creator of brew, tea is a standalone, binary download for all platforms that puts the entire open source ecosystem at your fingertips. Casually and effortlessly use the latest and greatest or the oldest and most mature from any layer of any stack. Break down the silos between programming communities, throw together scripts that use entirely separate tools and languages and share them with the world with a simple one-liner.

**Why?**

Even though `ani-cli` doesn't have a lot of dependencies,
still, since it's a shell script, these dependencies need to be downloaded externally via different package managers which can be painful, sometimes.

`tea` solves it by doing it all by itself in an **isolted** environment. 

**How?**

One of the features of `tea` is that it [can run scripts in an interpeter mode](https://github.com/teaxyz/cli#tea-the-universal-interpreter)

So, only something like this is needed

```bash
#!/bin/sh

#---
#dependencies:
#  curl.se: ^7.8
#  openssl.org: ^1.1
#  ffmpeg.org: ^5.1
#  anything-else: ^999
#---

# The rest of the script goes here...
```

It's just a comment that won't break other use-cases.

**Conclusion**

At the end of the day, all you'd need to do to run `ani-cli` is

```bash
git clone https://github.com/pystardust/ani-cli

# add ani-cli to the PATH
# ...

# Let tea handle handle the rest
tea ani-cli
# Search Anime
# > 
```

**But!**

Unfortunately, `tea` is a new tool and doesn't include `axel`, `mpv` and `iina` in their [pantries](https://github.com/teaxyz/pantry.zero) yet.
Of course, it can be easily solved by making a PR to them.

So, *this* PR is a *draft* that only includes `curl`, `openssl` and `ffmpeg` with a hope that missing dependencies will be supported by the `tea` soon.

## Checklist

> I didn't change the script itself, so everything works, obviously

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,5 +1,12 @@
 #!/bin/sh
 
+#---
+#dependencies:
+#  curl.se: ^7.8
+#  openssl.org: ^1.1
+#  ffmpeg.org: ^5.1
+#---
+
 # License preamble at the end of the file
 # Version number
 VERSION="3.4.6"


justchokingaround:
wow wow, this is very interesting. will take a more in depth look at this a little later, tysm for sharing

justchokingaround (approved):

Issue #938, opened 2022-12-16
Fix manpage


werdahias:
Please move the manpage to ani-cli.1 and add a GPL-3 license to it.

Derisis13:
How would this change the packaging process if we wish to include the manpage?

PR #940, opened 2022-12-16
Fix manpage

medanisjbara:
# Fix manpage

Addresses issue #938.

## Type of change

- [X] Documentation update

## Description
untared ani-cli.1.gz using `tar xf ani-cli.1.gz` and changed the licence.


ani-cli.1
```diff
@@ -0,0 +1,101 @@
...
+Show summary of options.
+.TP
+\f[B]-d\f[R]
+download episode.
+.TP
+\f[B]-p\f[R] \[lq]download_dir\[rq]
+download episode to specified directory.
+.TP
+\f[B]-q\f[R] \[lq]quality\[rq]
+set video quality (best|worst|360|480|720|1080).
```
Derisis13, line 43: The short descriptions for a, p and q are a formatting nightmare and since no other argument has them, we should just drop them for good.
image

ani-cli.1
```diff
@@ -0,0 +1,101 @@
...
+delete history
+.TP
+\f[B]-U\f[R]
+fetch update from github
+.TP
+\f[B]-V\f[R]
+print version number and exit
+.TP
+\f[B]-f\f[R]
+select provider to scrape first
```
Derisis13, line 66: This one is just wrong, `-r` is used to determine which provider to scrape first (1-3) and `-f` is for using fzf as an UI. (as partially corrected in #942, however I'd prefer it being corrected here)

ani-cli.1
```diff
@@ -0,0 +1,101 @@
...
+\f[B]-q\f[R] \[lq]quality\[rq]
+set video quality (best|worst|360|480|720|1080).
+.TP
+\f[B]-v\f[R]
+use VLC as the media player
+.TP
+\f[B]-s\f[R]
+watch anime together with friends, using Syncplay (works with mpv only)
+.PP
+\f[B]-d\f[R] download episode
```
Derisis13, line 51: this needs a line break to be indented as the others:
image

ani-cli.1
```diff
@@ -0,0 +1,101 @@
+.\" Automatically generated by Pandoc 2.9.2.1
+.\"
+.TH "ANI-CLI (1) ani-cli" "" "\[lq]June 19 2022\[rq]" "" ""
```
Derisis13, line 3: If we update the manpage the least that should change is this date. Possibly the authors as well.

ani-cli.1
```diff
@@ -0,0 +1,101 @@
...
+.PP
+\f[B]ani-cli\f[R] [-s] [-v] [-q ] [-a ] [-d | -p ] []
+.PD 0
+.P
+.PD
+\f[B]ani-cli\f[R] [-s] [-v] [-q ] -c
+.PD 0
+.P
+.PD
+\f[B]ani-cli\f[R] -h | -D | -U | -V
```
Derisis13, line 18: The whole synopsis is just a mess. 
image
I don't really understand what is it's purpose. We should find it one or if we can't, just reduce it to the minimum

Derisis13 (changes requested):
Sorry for pushing all the previous author's mistakes on you, but it appears Pandoc 2.9.2.1 did a mediocre job.
Also the change from a .tar.gz archive to a plaintext file: how does it change the packaging process? What extra commands are required to include the manpage in the package?

parona-source:
> Also the change from a .tar.gz archive to a plaintext file: how does it change the packaging process? What extra commands are required to include the manpage in the package?

Man pages can be installed compressed or uncompressed. But having it uncompressed allows upstreams to decide on the compression method. For example Gentoo has variables with which the end user can decide the compressions as well as Fedora. So instead of catering to Debian policy of `gzip -9` this would allow packagers to avoid an extra step with uncompressing.

Also previously the man page wasnt compressed (despite filename) but was actually an uncompressed tarball.

ani-cli.1
medanisjbara: IMO a better description is better than dropping them. Especially `-a` since it is useful and can easily be confused with `-e` (which doesn't exist but resembles more the word "episode").
I rewrote the OPTIONS (following alphabetical order the same way `man ls` and `man mkdir` are laid out) and removed duplicate options.
image

I hope you reconsider wanting to drop them.

ani-cli.1
medanisjbara: I am not sure what changes should be introduced to authors.

medanisjbara:
I also had to remove duplicate options. Looking forward to your feedback @Derisis13

Derisis13 (approved):
Very nice, I like it a lot. I see you replaced the previous author's name and address with yours - I think it's OK since you almost completely rewrote the manpage, but are you sure you want to put your contact address out there?

Derisis13:
> > Also the change from a .tar.gz archive to a plaintext file: how does it change the packaging process? What extra commands are required to include the manpage in the package?
> 
> Man pages can be installed compressed or uncompressed (not packed as tars though, tar.gz would be incorrect). And having it uncompressed isnt a bad thing while allowing easier editing in the repo, as upstreams have differing policies on the compression method eitherway. For example Gentoo has variables with which the end user can decide the compressions as well as Fedora (Fedora and Gentoo tooling though is capable of uncompressing). So instead of catering to Debian policy of `gzip -9` this would allow packagers to avoid an extra step with uncompressing.
> 
> Also previously the man page wasnt compressed (despite filename) but was actually an uncompressed tar (which lead to Gentoo naive uncompressing to fail as it expected gzip and not a tar). Man program could parse it in essence by accident as if you tar a plaintext file you get out a plaintext file with extra noise like you can see here [#942 (review)](https://github.com/pystardust/ani-cli/pull/942#pullrequestreview-1224550437).

Thanks, I learnt something today. It makes sense now.

medanisjbara:
> Very nice, I like it a lot. I see you replaced the previous author's name and address with yours - I think it's OK since you almost completely rewrote the manpage, but are you sure you want to put your contact address out there?

I honestly put my email just because the previous author did. Nothing more, But I guess I'll remove it since it serves no practical purpose.

port19x:
Looks good, thank you for the contribution

PR #968, opened 2023-01-09
V4 steamdeck

Nannk:
## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

adds flag to use flatpak mpv

## Checklist

- [x] any anime playing
- [ ] bumped version
- [x] quality works
- [x] downloads work
- [ ] syncplay -s works

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -192,21 +192,13 @@ play_episode() {
 	get_episode_url
 	case "$player_function" in
 	debug) printf "All links:\n%s\nSelected link:\n%s\n" "$links" "$episode" ;;
-	mpv | iina) is_mpv_from_flatpak ;;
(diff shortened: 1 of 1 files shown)


port19x (approved):
`dep_ch` feels wrong.
Perhaps only check for flatpak and don't handle the case where the mpv flatpak isn't installed?
Not sure what is elegant in this case.

Nannk:
> `dep_ch` feels wrong. Perhaps only check for flatpak and don't handle the case where the mpv flatpak isn't installed? Not sure what is elegant in this case.

in that case the standart dep_ch would be fine
but it feels wrong to not check for flatpak mpv if we check for everything else

PR #1019, opened 2023-02-04
Ani-cli UX Changes

athuld:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

Hi guys just want to fix UX related issues
- the episodes are displayed to descending order which is usually not a good experience for users
- the menu should display the next episode option first followed by other relevant option and currently the quit option is displayed first after each episode play ends

PS: don't have syncplay so didn't test that.. apart from that everything is working

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [ ] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `--dub` both work
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -182,14 +182,14 @@ search_anime() {
 
 # get the episodes list of the selected anime
 episodes_list() {
-	wget -q -O - "https://allanime.site/anime/$*" -U "$agent" | sed 's|\\||g' | sed -nE "s|.*$mode\":\[([0-9.\",]*)\].*|\1|p" | sed 's|,|\n|g; s|"||g'
+	wget -q -O - "https://allanime.site/anime/$*" -U "$agent" | sed 's|\\||g' | sed -nE "s|.*$mode\":\[([0-9.\",]*)\].*|\1|p" | sed 's|,|\n|g; s|"||g' | sort -n -k 1
 }
 
 # PLAYING
 
 process_hist_entry() {
 	ep_list=$(episodes_list "$id")
-	ep_no=$(printf "%s" "$ep_list" | sed -n "/^${ep_no}$/{g;1!p;};h") 2>/dev/null
+	ep_no=$(printf "%s" "$ep_list" | sed -n "/^${ep_no}$/{n;p;}") 2>/dev/null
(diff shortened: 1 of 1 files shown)


port19x (commented):
I agree that `next` should be the default option in the interactive menu.
I'm however uncertain that an ascending order makes more sense.
Viewers of airing shows for example might only be interested in the latest, tho since those can and should use `-c` that still leaves your stance valid.
I'll make a poll

athuld:
As with the case of new users seeing the last episode of the anime as the first option to select might not be a good experience and even though you can scroll up the fzf to reach the first episode, most users don't know this and by default scrolls down. This is usually not an ideal way to navigate the episodes in my opinion. Given an ordered list new users can easy go through the episode and viewers of airing shows usually uses `-c `option or since most anime are usuallyy released in seasons .. making the order ascending seems to be the better choice

port19x:
> As with the case of new users seeing the last episode of the anime as the first option to select might not be a good experience and even though you can scroll up the fzf to reach the first episode, most users don't know this and by default scrolls down. This is usually not an ideal way to navigate the episodes in my opinion. Given an ordered list new users can easy go through the episode and viewers of airing shows usually uses `-c `option or since most anime are usuallyy released in seasons .. making the order ascending seems to be the better choice

I always assumed the default behaviour to be hitting `1` `RET`
But yeah, makes sense I guess now that you put it like that

port19x:
Alright then

port19x (approved):

port19x:
My god, why was the executable bit removed, do you have brain damage?
Rip master

justchokingaround:
xdddddd

athuld:
> My god, why was the executable bit removed, do you have brain damage? Rip master

My bad sorry I was testing it with user permission..I might have never noticed it..sorry

athuld:
https://github.com/pystardust/ani-cli/pull/1019/commits/419859c71134501c0fa67692f660ec14f62a4bde

@port19x  this is what caused the bit change.It was not my mistake...next time instead of blindly calling someone brain dead you should check the facts

PR #1020, opened 2023-02-05
search field changed into while loop

r4vanan:
i don't want to turn back after entering into ani-cli tool so i done this. if you like too then pull it else leave it thanks

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -355,15 +355,14 @@ history)
 			printf "Search anime: " && read -r query
 		else
 			query=$(: | external_menu "" "Search anime:")
-		fi
-	done
-	query=$(printf "%s" "$query" | sed "s| |%20|g")
-	anime_list=$(search_anime "$query")
-	while [ -z "$anime_list" ]; do
-		printf "Search anime: " && read -r query
-		search_anime "$query"
-		query=$(printf "%s" "$query" | sed "s| |%20|g")
-		anime_list=$(search_anime "$query")
+		fi 
+		while [ -z $anime_list ]; do
+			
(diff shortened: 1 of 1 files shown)


port19x:
It doesn't quite behave as expected but I agree with that UX

ani-cli
```diff
@@ -352,14 +352,20 @@ history)
 *)
 	while [ -z "$query" ]; do
 		if [ "$use_external_menu" = "0" ]; then
+			clear
```
Derisis13, line 355: Why clear here?

ani-cli
```diff
@@ -352,14 +352,20 @@ history)
...
 			printf "Search anime: " && read -r query
 		else
 			query=$(: | external_menu "" "Search anime:")
 		fi
 	done
 	query=$(printf "%s" "$query" | sed "s| |%20|g")
 	anime_list=$(search_anime "$query")
-	[ -z "$anime_list" ] && die "No results found!"
+	while [ -z "$anime_list" ]; do
+		clear
```
Derisis13, line 364: There is an escape sequence that sets the cursor to the previous line and deletes it. Use that, not clear.

ani-cli
```diff
@@ -352,14 +352,20 @@ history)
...
+		printf "Search anime: " && read -r query
+		search_anime "$query"
```
Derisis13, line 366: The regex for replacing spaces is missing from here

Derisis13 (changes requested):
Lots of repeated lines, there should be a way to solve it without that.

PR #1023, opened 2023-02-06
search field changed into while loop

r4vanan:
i don't want kill that process while for searching the anime so i made a while loop for that.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -155,7 +155,14 @@ select_quality() {
 	[ -z "$result" ] && printf "Specified quality not found, defaulting to best" 1>&2 && result=$(printf "%s" "$links" | head -n1)
 	printf "%s" "$result" | cut -d'>' -f2
 }
-
+search_loop(){
+	while [ -z $anime_list ]; do
+		printf "Search anime: " && read -r query
+		query=$(printf "%s" "$query" | sed "s| |%20|g")
+		search_anime "$query"
+		anime_list=$(search_anime "$query")
+	done
+}
 # gets embed urls, collects direct links into provider files, selects one with desired quality into $episode
 get_episode_url() {
 	# get the embed urls of the selected episode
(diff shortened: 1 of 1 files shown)


71zenith:
appease the linter

r4vanan:
> appease the linter

I can't understand what you mean sir

r4vanan:
I think now it's ok

RaynardGerraldo:
searching for anime as argument doesnt work with this, ./ani-cli naruto just prompts me to search anime again

ani-cli
```diff
@@ -352,14 +359,12 @@ history)
...
-			printf "Search anime: " && read -r query
+			search_loop
 		else
 			query=$(: | external_menu "" "Search anime:")
 		fi
 	done
-	query=$(printf "%s" "$query" | sed "s| |%20|g")
-	anime_list=$(search_anime "$query")
-	[ -z "$anime_list" ] && die "No results found!"
+	search_loop
```
port19x, line 367: this would break external menu compatibility

port19x (changes requested):

r4vanan:
i think now `./ani-cli naruto` it can possible

r4vanan:
now it can work with rofi

RaynardGerraldo:
> i don't want kill that process while for searching the anime so i made a while loop for that.

can you explain what exactly this means and what it improves? because I dont see a difference, you can show some screenshots

r4vanan:
if i search the anime name. if the result is not found means it kill the process i dont like that process. so i changed that and do the loop thing. like if you enter the wrong or invalid name it will ask the name again. it will work in rofi too. no matter how many time you entered wrong name. if the result not found means it will ask the name again

RaynardGerraldo:
> if i search the anime name. if the result is not found means it kill the process i dont like that process. so i changed that and do the loop thing. like if you enter the wrong or invalid name it will ask the name again. it will work in rofi too. no matter how many time you entered wrong name. if the result not found means it will ask the name again

ok, got it

port19x:
This is straightup poorly implemented. Not sure if you can refactor it, or if it needs to be this branchy in order to work

r4vanan:
You can close it. Because I like this thing so I modified like this

PR #1076, opened 2023-04-01
added debian instructions

werdahias:
# Pull Request Template

## Type of change

- [ x] Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,12 +69,9 @@ Otherwise you're likely to see an error like the following: ` "/usr/bin/ani-cli:
 
 ### Linux
 
-#### Debian
+#### Debian unstable
 
 ```sh
-curl -s https://Wiener234.github.io/ani-cli-ppa/KEY.gpg | sudo tee /etc/apt/trusted.gpg.d/ani-cli.asc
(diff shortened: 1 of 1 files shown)


justchokingaround:
might as well just remove the whole section at this point

werdahias:
Why?

justchokingaround:
oh sorry, i didn't see the line with the apt command. this is actually great news. can someone test it on debian, to make sure it works?

werdahias:
You'd need debian unstable to test it. FWIW it works fine for me

Wiener234:
I would suggest leaving the part about uninstalling the ppa in the Readme for some time to make sure people who had the ppa remove it first before trying to install from the official servers.

port19x:
> might as well just remove the whole section at this point

Do you mean the distro specifics in general?
That might be a discussion worth having for sure.
We're in more and more repos, links to the right places may be superior and less error prone than telling everyone what command to copy paste

nightwolf3140 (commented):
`can someone test it on debian, to make sure it works?`
Changes seem great but cant confirm the package (im on arch)

port19x:
Yeah, there is no way it doesn't work on Debian Unstable

port19x (approved):

PR #1080, opened 2023-04-06
docs: prune and reorder homies

### LINKED ISSUE #1071: tv-cli link is not working
**Describe the bug**
The [tv-cli](https://github.com/Spaxly/tv-cli) link leads to a GitHub error page, all the other links are working fine.

**Screenshots (if applicable; you can just drag the image onto github)**
image

port19x:
# Pull Request Template

closes #1071 

## Type of change

- [ ] Bug fix
- [ ] Feature
- [X] Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,4 +374,4 @@ flatpak uninstall io.mpv.Mpv
 * [lobster](https://github.com/justchokingaround/lobster): Watch movies and series from the terminal (Shell)
 * [mov-cli](https://github.com/mov-cli/mov-cli): Watch movies/shows in the cli (Python/Shell)
 * [dra-cla](https://github.com/CoolnsX/dra-cla): ani-cli equivalent for korean dramas (Shell)
-* [redqu](https://github.com/port19x/redqu):  A media centric reddit client 
+* [redqu](https://github.com/port19x/redqu):  A media centric reddit client (Clojure)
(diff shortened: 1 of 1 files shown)


justchokingaround (approved):

Derisis13:
Since we're doing this, do we want to extract the massive list of installations/uninstallations into a separate `installing.md` that is hyperlinked in the readme?
It would be an extra step for new users, for a more streamlined readme, but we're at the point where the readme is almost as large as the whole script

justchokingaround:
another alternative would be for use to write a proper wiki page

port19x:
> another alternative would be for use to write a proper wiki page

I'm against using github wiki for this, since it's not version controllable

port19x:
> Since we're doing this, do we want to extract the massive list of installations/uninstallations into a separate `installing.md` that is hyperlinked in the readme? It would be an extra step for new users, for a more streamlined readme, but we're at the point where the readme is almost as large as the whole script (373 lines and 430 lines)

I'd prefer just omitting the sections listing native packages and how to install them, especially if they appear in the repology badge.
I trust that especially linux users are capable of installing a program from their native package manager.
We don't have to tell them to `sudo apt install ani-cli`

port19x:
Anyway I believe this should be properly discussed in another issue that I'll open

PR #1090, opened 2023-04-17
Added option to use the jellyfin naming format. This requires adding …

Lop010:
…the -j --jellyfin and -Se --season option. A complete command using this new feature would be "ani-cli --jellyfin True --season 2 -d "attack on titan"

# Pull Request Template

## Type of change

- [ ] Bug fix
- [X] Feature
- [ ] Documentation update

## Description

So recently i've started to use Jellyfin but i was mildly annoyed at the formatting so i made a python script. Now i have implemented this into ani-cli. This allows you to use the --jellyfin and --season to turn the messy formatting you normally have into something jellyfin picks up immediately   

As you can see in the code only changes the download function, --help display and the "argument handler" so it wont cause any unforeseen bugs.

## Checklist

- [X] any anime playing
- [ ] bumped version
---
- [X] `-d` downloads work
- [X] `-s` syncplay works
- [X] `-q` quality works
- [X] `-v` vlc works
- [X] `-e` select episode works
- [X] `-S` select index works
- [X] `-r` range selection works
- [X] `--dub` both work
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

- Non-whole episodes: Tensei shitara slime datta ken (ep. 24.5, ep. 24.9)  (This does not work currently im not good enough to solve this)


ani-cli
```diff
@@ -63,12 +63,18 @@ help_info() {
         play dubbed version
       -U, --update
         Update the script
+      -Se --season
+        Select season (Needed for the -j --jellyfin option)
```
port19x, line 67: Bad idea. `-Se` will get parsed by getopts as `-S` (select-nth) and `-e` (episode)

ani-cli
```diff
@@ -63,12 +63,18 @@ help_info() {
...
+      -j --jellyfin
```
port19x, line 68: Don't make it take True/False.
Absence means false. Presence means true.

ani-cli
```diff
@@ -316,7 +339,20 @@ while [ $# -gt 0 ]; do
...
+        -Se | --season)
+            [ $# -lt 2 ] && die "missing argument!"
+            seasonnum="$2"
+            shift
+            ;;        
+        
         -c | --continue) search=history ;;
+        
+        --j | --jellyfin)
+            [ $# -lt 2 ] && die "missing argument!"
```
port19x, line 351: As noted previously, presence means true, absence means false.
Also it's `-j` not `--j` if at all

ani-cli
```diff
@@ -211,10 +217,27 @@ update_history() {
...
-        *) aria2c --check-certificate=false --continue --summary-interval=0 -x 16 -s 16 "$1" --dir="$download_dir" -o "$2.mp4" --download-result=hide ;;
+        *)
+            if [ "$jellyfin" = "True" ]; then
+                if [ ${#ep_no} -eq 3 ]; then
+                    epstr="E$ep_no"
+                elif [ ${#ep_no} -eq 2 ]; then
+                    epstr="E0$ep_no"
+                elif [ ${#ep_no} -eq 1 ]; then
+                    epstr="E00$ep_no"
+                fi
```
port19x, line 231: This is bad code, but could work for detecting the number of eps I guess.
Zero padding is usually done differently in shell.
Also, this would break for really long anime, like one piece, with >1k episodes

ani-cli
```diff
@@ -211,10 +217,27 @@ update_history() {
     mv "${histfile}.new" "$histfile"
 }
 
+
 download() {
     case $1 in
-        *m3u8*) ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4" ;;
-        *) aria2c --check-certificate=false --continue --summary-interval=0 -x 16 -s 16 "$1" --dir="$download_dir" -o "$2.mp4" --download-result=hide ;;
```
port19x, line 217: You probably broke m3u8 compatibility in your code.

port19x (changes requested):

trueauracoral:
*This should probably be an issue but this is related*

What if someone like me is using kodi or maybe plex, it requires a different file format than jellyfin does. It would be cool if ani-cli had an argument (-o/--output) to format however you like the output file name. An example could be:

```sh
ani-cli -o "S%(season_number)E%(episode_number)s.%(ext)s"
```

Youtube-dl/yt-dlp has this option with -o and output template they document it in their readme https://github.com/yt-dlp/yt-dlp#output-template

It uses python's string formatting. Possibly bash some equivalent to this instead of using sed/regex. Also somehow padding would need to be implemented. 

This is very difficult to implement but I could be wrong.

CoolnsX:
> _This should probably be an issue but this is related_
> 
> What if someone like me is using kodi or maybe plex, it requires a different file format than jellyfin does. It would be cool if ani-cli had an argument (-o/--output) to format however you like the output file name. An example could be:
> 
> ```shell
> ani-cli -o "S%(season_number)E%(episode_number)s.%(ext)s"
> ```
> 
> Youtube-dl/yt-dlp has this option with -o and output template they document it in their readme https://github.com/yt-dlp/yt-dlp#output-template
> 
> It uses python's string formatting. Possibly bash some equivalent to this instead of using sed/regex. Also somehow padding would need to be implemented.
> 
> This is very difficult to implement but I could be wrong.

Actually this is the best way, rather than using args per self-hosted streaming service..

port19x:
> _This should probably be an issue but this is related_
> 
> What if someone like me is using kodi or maybe plex, it requires a different file format than jellyfin does. It would be cool if ani-cli had an argument (-o/--output) to format however you like the output file name. An example could be:
> 
> ```shell
> ani-cli -o "S%(season_number)E%(episode_number)s.%(ext)s"
> ```
> 
> Youtube-dl/yt-dlp has this option with -o and output template they document it in their readme [yt-dlp/yt-dlp#output-template](https://github.com/yt-dlp/yt-dlp#output-template)
> 
> It uses python's string formatting. Possibly bash some equivalent to this instead of using sed/regex. Also somehow padding would need to be implemented.
> 
> This is very difficult to implement but I could be wrong.

My problem with this and with the PR itself is that bulk rename tools are plentiful these days and I don't see how someone that knows how to define a correct output string can't bulk rename downloaded files themselves.
This (-o and the PR) seems out of scope

trueauracoral:
> My problem with this and with the PR itself is that bulk rename tools are plentiful these days and I don't see how someone that knows how to define a correct output string can't bulk rename downloaded files themselves. This (-o and the PR) seems out of scope

Majority of bulk renaming tools I know of require API keys. It's so much easier if the files are downloaded with filenames/file format you need from the start.

PR #1092, opened 2023-04-20
Add --output argument :D

Lop010:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

Added a --output argument. Can be used similarly to the same argument from yt-dlp

USAGE:
 - Supports allanime_title,  (%AT), mode (sub, dub)  (%M), ep-num  (%E), seasonnum (if its provided in --season NUM)  (%S)
 - Example of use can be '%AT%M-%E-%S'"
 - The defult download format can be recreated with "%AT%E-%M"

Also added -e --season to manually select a season number as i couldnt find a variable that was there for that. Also needs a better argument name then -e
## Checklist

- [ x ] any anime playing
- [ ] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--dub` both work
---
- [ ] Readme is up to date
- [ ] Man page is up to date

### This formatting was used for testing "%AT%M-%E-%S"

- The safe bet: One Piece | This works (Downloaded a few episodes, i think 10)
- Episode 0: Saenai Heroine no Sodatekata ♭ | This works
- Unicode: Saenai Heroine no Sodatekata ♭ |  This works
- Non-whole episodes: Tensei shitara slime datta ken (ep. 24.5, ep. 24.9) | This works


ani-cli
```diff
@@ -63,12 +63,19 @@ help_info() {
         play dubbed version
       -U, --update
         Update the script
+      -Se --season
+        Select season (Needed for the -j --jellyfin option)
+      -o --output
+        Supports allanime_title  (%AT), mode (sub, dub)  (%M), ep-num  (%E), seasonnum (if its provided in --season NUM)  (%S)
+        Example of use can be 'Ep%ES%S-%m'
+        The defult way is \"%AT%E-%M\"
```
justchokingaround, line 71: default*

ani-cli
```diff
@@ -63,12 +63,19 @@ help_info() {
         play dubbed version
       -U, --update
         Update the script
+      -Se --season
```
justchokingaround, line 66: :skull:

ani-cli
```diff
@@ -213,8 +220,28 @@ update_history() {
 
 download() {
     case $1 in
-        *m3u8*) ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4" ;;
-        *) aria2c --check-certificate=false --continue --summary-interval=0 -x 16 -s 16 "$1" --dir="$download_dir" -o "$2.mp4" --download-result=hide ;;
+        *m3u8*)
+            if [ -n "$output" ]; then
+                nameformat=$("echo "$output" | sed 's/%E/$ep_no/' | sed 's/%E/$seasonnum/'") # Replace the
```
justchokingaround, line 225: - quotes should be outside of the brackets `"(..)"` instead of `("...")`
- use printf instead of echo
- chain sed commands using `;` at the end of one of them -> `sed 's/%E/$ep_no/;s/%E/$seasonnum/'`

Lop010:
Pretty sure i've fixed all the thing's you pointed out in the new commit

ani-cli
```diff
@@ -63,12 +63,19 @@ help_info() {
         play dubbed version
       -U, --update
         Update the script
+      -e --season
```
port19x, line 66: `-e` is already take for `--episode`.
I'm also not sure season selection work well if at all, since many anime don't have numbered seasons, but rather differently named continuations.
If we can have `-o` without `-e`, that would be prefered.
Also consider `-f` for `format` over `-o` for `output`, tho I have no strong preference on this

port19x (changes requested):

ani-cli
Lop010: Sorry didnt notice that -e it was taken. And i dont really have a preference either for the -f or -o thing. If people want it i guess i could change it
I will change the -e but not sure what i can change it with

Lop010:
Wops sorry bad at doing git stuff

Lop010:
Does anybody know why the check failed or does it not matter?

Lop010:
I'll have to be honest here, I have no idea what that means or how to fix it.

justchokingaround:
just change it to this:
```sh
      -f --format
        Supports allanime_title  (%%AT), mode (sub, dub)  (%%M), ep-num  (%%E), seasonnum (if its provided in --season NUM)  (%%S)
        Example of use can be 'Ep%%ES%%S-%%m'
        The default way is \"%%AT%%E-%%M\"
    Some example usages:
      %s -q 720p banana fish
      %s -d -e 2 cyberpunk edgerunners
      %s --vlc cyberpunk edgerunners -q 1080p -e 4
      %s blue lock -e 5-6
      %s -e \"5 6\" blue lock
      %s -d -o \"S%%SE%%E-%%M\" --dub --season 1 \"attack on titan\"
    \n" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}" "${0##*/}"
```

Lop010:
Thanks very much will do when I get home in a few hours

Lop010:
@port19x Could you close the change requested as it is now resolved

Lop010:
Wait so why was this closed?

PR #1115, opened 2023-05-09
fix: account for syncplay in either default location or windows path

### LINKED ISSUE #1114: On Windows, ani-cli cannot detect my installation of Syncplay despite Syncplay being on the path
Version: 4.2.7
OS: Windows 10
Shell: Git Bash (from Git for Windows)
Anime: (Not Applicable)

**Describe the bug**
Downloading is broken.
It says something about an unsupported protocol, see screenshot.

**Steps To Reproduce**

1. Install ani-cli through scoop (`scoop install ani-cli`) or clone the repository.
2. Install any other dependencies through scoop:
    ```
    scoop bucket add extras
    scoop install git fzf mpv aria2 syncplay
    ```
4. Run ani-cli with "-s": `ani-cli -s`
5. Observer error:
   ```
   $ ./ani-cli -s
   Checking dependencies...
   Program "/c/Program Files (x86)/Syncplay/Syncplay.exe" not found. Please install it.
   ```

**Expected behavior**
Since Syncplay is added to the path via scoop, ani-cli should find and use it.

**Additional context**
It looks like the ani-cli script is hardcoded to find Syncplay at "C:\Program Files (x86)/Syncplay/Syncplay.exe", instead of finding it through the path. This may be because the Syncplay installer defaults to installing there, as well as the choco package manager.

ykhan21:
…or on the path for Windows

# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description
This fixes #1114.
This fix accounts for Syncplay being at either the default location (used by the Syncplay installer and the chocolatey package manager) or on the path.
If Syncplay doesn't exist in the default location, it will set `player_function` to "syncplay.exe", expecting it to come from the path.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--dub` both work
---
- [x] Readme is up to date
- [x] Man page is up to date

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.2.9"
+version_number="4.3.0"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.2.7"
+version_number="4.2.8"
 
 # UI
 
@@ -323,7 +323,10 @@ while [ $# -gt 0 ]; do
         -s | --syncplay)
             case "$(uname -s)" in
                 Darwin*) player_function="/Applications/Syncplay.app/Contents/MacOS/syncplay" ;;
-                MINGW* | *Msys) player_function="/c/Program Files (x86)/Syncplay/Syncplay.exe" ;;
+                MINGW* | *Msys)
+                    player_function="/c/Program Files (x86)/Syncplay/Syncplay.exe"
+                    command -v "$player_function" >/dev/null || player_function="syncplay.exe"
+                    ;;
                 *) player_function="syncplay" ;;
             esac
             ;;
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.2.8"
+version_number="4.2.9"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -324,8 +324,8 @@ while [ $# -gt 0 ]; do
             case "$(uname -s)" in
                 Darwin*) player_function="/Applications/Syncplay.app/Contents/MacOS/syncplay" ;;
                 MINGW* | *Msys)
-                    player_function="/c/Program Files (x86)/Syncplay/Syncplay.exe"
-                    command -v "$player_function" >/dev/null || player_function="syncplay.exe"
+                    export PATH=$PATH:"/c/Program Files (x86)/Syncplay/"
+                    player_function="syncplay.exe"
                     ;;
                 *) player_function="syncplay" ;;
             esac


ani-cli
```diff
@@ -323,7 +323,10 @@ while [ $# -gt 0 ]; do
         -s | --syncplay)
             case "$(uname -s)" in
                 Darwin*) player_function="/Applications/Syncplay.app/Contents/MacOS/syncplay" ;;
-                MINGW* | *Msys) player_function="/c/Program Files (x86)/Syncplay/Syncplay.exe" ;;
+                MINGW* | *Msys)
+                    player_function="/c/Program Files (x86)/Syncplay/Syncplay.exe"
+                    command -v "$player_function" >/dev/null || player_function="syncplay.exe"
+                    ;;
```
Derisis13, line 329: I like the approach, however I'd do it in reverse: first check if it's in the path then program files (x86).

Derisis13 (changes requested):
See comment

ani-cli
ykhan21: I thought of that, but one thing to note about this approach the error message from `dep_ch()` will give will simply be `Program "syncplay.exe" not found. Please install it.` So, a user would install Syncplay either through the normal installer / choco or through scoop.

If I reverse it, the error message from `dep_ch()` will be `Program "/c/Program Files (x86)/Syncplay/Syncplay.exe" not found. Please install it.` This will be inconsistent with the other error messages. For example, if fzf is not found the error message will be `Program "fzf" not found. Please install it.`

Another thing to note is that the "default" location for Syncplay may be in `/c/Program Files (x86)`.

Would you still like me to reverse it?

Derisis13: Oh, that makes sense... Another question is why is it not a good idea to add `Program Files (x86)` to path (because if we can get away with just that, then I'd rather have it that way...)

ykhan21: It could simply be:
```
MINGW* | *Msys) player_function="syncplay.exe" ;;
```
But Windows users would have to manually add it to the path if they downloaded it through means other than the scoop package manager. Then we might have issues like "Why can't ani-cli detect Syncplay"?
We could go with that since we're already asking them to do a lot of setup (like getting a new terminal and adding git bash as a profile it).

ykhan21: Or do you mean something like this:
```
MINGW* | *Msys)
    export PATH=$PATH:"/c/Program Files (x86)/Syncplay/"
    player_function="syncplay.exe"
    ;;
```
This should cover for both download locations. It works on my machine despite Syncplay.exe being capitalized in `/c/Program Files (x86)/Syncplay/`.

port19x:
This will be the 4.3 once merged later today

port19x:
flippin heck, why would I need the command line to resolve a single line merge confligt github

port19x:
*cracks knuckles*

PR #1130, opened 2023-05-17
docs: Add iOS uninstall instructions

Lockl00p:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [X] Documentation update

## Description

*ramble here*

## Checklist

- [X] any anime playing
- [X] bumped version
---
- [X] next, prev and replay work
- [X] `-c` history and continue work
- [X] `-d` downloads work
- [X] `-s` syncplay works
- [X] `-q` quality works
- [X] `-v` vlc works
- [X] `-e` select episode works
- [X] `-S` select index works
- [X] `-r` range selection works
- [X] `--dub` and regular (sub) mode both work
- [X] all providers return links (not necessarily on a single anime, use debug mode to confirm)
---
- [X] `-h` help info is up to date
- [X] Readme is up to date
- [X] Man page is up to date

## Additional Testcases

- The safe bet: One Piece
- Episode 0: Saenai Heroine no Sodatekata ♭
- Unicode: Saenai Heroine no Sodatekata ♭
- Non-whole episodes: Tensei shitara slime datta ken (ep. 24.5, ep. 24.9)

PR #1131, opened 2023-05-20
test: create inverse jobs to catch required conditionals

port19x:
Fix for [conditional but required jobs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

### DIFF
--- a/.github/workflows/inverse-ani.yml
+++ b/.github/workflows/inverse-ani.yml
@@ -8,14 +8,17 @@ on:
       - "**ani-cli"
 jobs:
   sh-checker:
+    name: Shellcheck + Shfmt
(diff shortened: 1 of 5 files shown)


port19x:
Did I name the actions wrong?!

Derisis13:
I get the gist of what you did, how do I test that it works though?

port19x:
> I get the gist of what you did, how do I test that it works though?

You could check the actions output and also check out the behavior on #1130 
I also linked to the relevant docs, as I followed it pretty closely.

I don't think you can check on your local machine

PR #1154, opened 2023-06-17
added the '--save-position-on-quit' to continue from where the user h…

c2700:
…ad left off on said episode

# Pull Request Template

## Type of change

- [ ] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

*ramble here*

## Checklist

- [ ] any anime playing
- [ ] bumped version
---
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-S` select index works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -252,7 +252,7 @@ play_episode() {
     [ -z "$episode" ] && get_episode_url
     case "$player_function" in
         debug) printf "All links:\n%s\nSelected link:\n%s\n" "$links" "$episode" ;;
-        mpv*) nohup "$player_function" --force-media-title="${allanime_title}episode-${ep_no}-${mode}" "$episode" >/dev/null 2>&1 & ;;
+        mpv*) nohup "$player_function" --save-position-on-quit --force-media-title="${allanime_title}episode-${ep_no}-${mode}" "$episode" >/dev/null 2>&1 & ;;
         android_mpv) nohup am start --user 0 -a android.intent.action.VIEW -d "$episode" -n is.xyz.mpv/.MPVActivity >/dev/null 2>&1 & ;;
         android_vlc) nohup am start --user 0 -a android.intent.action.VIEW -d "$episode" -n org.videolan.vlc/org.videolan.vlc.gui.video.VideoPlayerActivity -e "title" "${allanime_title}episode-${ep_no}-${mode}" >/dev/null 2>&1 & ;;
         iina) nohup "$player_function" --no-stdin --keep-running --mpv-force-media-title="${allanime_title}episode-${ep_no}-${mode}" "$episode" >/dev/null 2>&1 & ;;


Derisis13 (changes requested):
If only it was this easy. But the different providers' links are treated as separate media, so if you get a different one the next time, it won't work. And the links don't live forever, which means that once they expire it'll be a separate media, won't work as intended.

Also there's the ambiguity of whether you just want to skip the end credits or you couldn't finish the episode in your lunch break. 

All of this can be probably resolved, but it needs some thinking

justchokingaround:
mpv has a keybind by default to do this, you can press shift+q instead of just q to quit. you can also set the option to save on quit by default in your mpv.conf

c2700:
> mpv has a keybind by default to do this, you can press shift+q instead of just q to quit. you can also set the option to save on quit by default in your mpv.conf

Learned something new today :)

c2700:
> If only it was this easy. But the different providers' links are treated as separate media, so if you get a different one the next time, it won't work. And the links don't live forever, which means that once they expire it'll be a separate media, won't work as intended.
> 
> Also there's the ambiguity of whether you just want to skip the end credits or you couldn't finish the episode in your lunch break.
> 
> All of this can be probably resolved, but it needs some thinking

My bad. I opened this PR cuz it worked for as intended on all the episodes that I have been watching & have watched & skipping the outri & intro was something that didn't cross my mind. But to skip the outro & intro is it possible for sponsorblock to be used here? Sure the sponsorblock database may need to be updated every now & then but to reduce the work all the info of all available intros & outros could be scraped from vimeo, youtube or dailymotion & that could be put into the sponsorblock's database which will be applied to all the episodes that have the same intro & outro. This is just one way I could come up with skipping segements cuz I did see mpv being shown as a 3rd party port in the sponsorblock's site

justchokingaround:
@c2700 
- there is already an mpv extension to do that: https://github.com/rui-ddc/skip-intro (it's set to the TAB keybind by default, but you can change that). there is also a sponsorblock extension, but that's for youtube, not for anime. skip-intro has never failed me once for anime in ~1.5 years of using it (if u're interested in this, u can check my mpv config: https://github.com/justchokingaround/dotfiles/tree/main/mpv)
- the reason why your code in the pr worked in your testing, is because the links don't expire immediately, so you got lucky, but it is not a feasible approach in general
- the feature that you want, where the progress is saved to the exact second, can be implemented through mpv's socket and socat, here is an example of how that can be done: https://github.com/justchokingaround/lobster/blob/main/lobster.sh#L400-L417
however this feature seems kinda like bloat in ani-cli's case, and it would require rewriting the history implementation + an additional dependency

c2700:
will check them all out. thank you again :)

justchokingaround:
this part would require a rewrite: https://github.com/pystardust/ani-cli/blob/master/ani-cli#L220-L234, this part too: https://github.com/pystardust/ani-cli/blob/master/ani-cli#L398-L412, the play function would require many additional checks, and it would add many lines of code, especially since there is a need for an implementation with socat and without it, since it would be an optional dependency

c2700:
okay, now that you've pointed me over there I see how much of a change it will need. mind if I implement the changes? (As long as the position tracker alone isn't bloat in ani-cli's case i.e.)

justchokingaround:
you can implement it if ubwish, but it's unlikely that we will merge it into ani-cli. socat must be an optional dependency, because it doesn't exist on windows, so u must also have coverage for the case where u just play videos, without using socat and saving position

c2700:
okay, will implementing the necessary changes then

justchokingaround:
what?

c2700:
As in, I'll be making the changes necessary to implement the "start from where I left off" if you're okay with it

justchokingaround:
sure, u can hmu if u got any questions or need help. i'm personally in favor of this feature, but i dont think thats the case for all maintainers

c2700:
sure, will do :)

PR #1155, opened 2023-06-17
feat: refactor code and optional hls script

CoolnsX:
# Pull Request Template

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

*ramble here*

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -406,13 +406,14 @@ case "$search" in
         tput cuu1 && tput el
         ;;
     *)
-        while [ -z "$query" ]; do
-            if [ "$use_external_menu" = "0" ]; then
+        if [ "$use_external_menu" = "0" ]; then
+            while [ -z "$query" ]; do
                 printf "Search anime: " && read -r query
-            else
-                query=$(: | external_menu "" "Search anime: ")
-            fi
-        done
+            done
+        else
+            query=$(printf "" | external_menu "" "Search anime: ")
+            [ -z "$query" ] && exit 1
+        fi
         query=$(printf "%s" "$query" | sed "s| |+|g")
         anime_list=$(search_anime "$query")
         [ -z "$anime_list" ] && die "No results found!"
--- a/ani-cli
+++ b/ani-cli
@@ -230,14 +230,16 @@ update_history() {
 
 download() {
     case $1 in
-        *m3u8*)	if command -v "hls";then
-			hls -n 300 -ro "$download_dir/$2" "$1"
-		else
-			ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4"
-		fi
-		;;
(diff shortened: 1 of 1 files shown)


CoolnsX:
finally!!!

ani-cli
```diff
@@ -128,7 +125,7 @@ get_links() {
             ;;
         *) [ -n "$episode_link" ] && printf "%s\n" "$episode_link" ;;
     esac
-    printf "Fetching %s Links\n" "$provider_name" 1>&2
+    printf "\033[1;32m %s\033[0m Links Fetched\n" "$provider_name" 1>&2
```
Derisis13, line 128: Good one. But now that you opened tha UI fixin box, I'm not stopping until the deleting of lines is consistent

ani-cli
```diff
@@ -235,16 +230,16 @@ update_history() {
...
     case $1 in
-        *m3u8*) ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4" ;;
-        *) if curl -s -m 5 "$1" | grep -q '^#EXTM3U'; then
-            ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4"
-        else
-            if uname -a | grep -q "ish"; then
-                curl --output-dir "$download_dir" -o "$2.mp4" "$1"
+        *m3u8*)
+            if command -v "hls"; then
+                hls -n 300 -ro "$download_dir/$2" "$1"
```
Derisis13, line 235: I wonder if there's a way to do this with `dep_ch` and `player_fn`. Especially that in this state not having hls on ish would lead to an undefined state. (I suppose the check was in place because ffmpeg is unavailable on ish?)

Derisis13 (commented):
I have some questions, but I think that the inclusion of your hls script is the closest thing we got to a plugin. This is why I think we shouldn't add it in the main script, but create a plugin system and add hls there as an example. I imagine it like *sh.rc, we just source it before/after we set up variables.

ani-cli
justchokingaround: wouldn't just adding `tput clear` above this line be sufficient for that?

ani-cli
CoolnsX: check was in place in case someone didn't install the script, and yes you can use dep_ch I was writing the code when body temp was high, so I didn't think about that

port19x:
I'll take a closer look later today, possibly tomorrow

port19x (approved):

candrapersada:
and how to auto clean `${HOME}/.cache/hls-temp` after the download is complete?

PR #1165, opened 2023-06-30
fix: hex decryption

CoolnsX:
# Pull Request Template

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

*ramble here*

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
---
- [ ] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -317,11 +317,7 @@ use_external_menu="${ANI_CLI_EXTERNAL_MENU:-0}"
 [ "$use_external_menu" = "0" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-m"}"
 [ "$use_external_menu" = "1" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-multi-select"}"
 cache_dir="${ANI_CLI_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/ani-cli}"
-if [ ! -d "$cache_dir" ]; then
-    mkdir -p "$cache_dir"
-else
-    rm -f "$cache_dir/*"
-fi
+[ ! -d "$cache_dir" ] && mkdir -p "$cache_dir"
 hist_dir="${ANI_CLI_HIST_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/ani-cli}"
 [ ! -d "$hist_dir" ] && mkdir -p "$hist_dir"
 histfile="$hist_dir/ani-hsts"
--- a/ani-cli
+++ b/ani-cli
@@ -228,7 +228,7 @@ update_history() {
 download() {
     case $1 in
         *m3u8*)
-            if command -v "hls"; then
+            if command -v "hls" >/dev/null; then
                 hls -ro "$download_dir/$2" "$1"
             else
                 ffmpeg -loglevel error -stats -i "$1" -c copy "$download_dir/$2.mp4"
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.4.6"
+version_number="4.5.0"
 
 # UI
 
@@ -322,7 +322,11 @@ use_external_menu="${ANI_CLI_EXTERNAL_MENU:-0}"
 [ "$use_external_menu" = "0" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-m"}"
 [ "$use_external_menu" = "1" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-multi-select"}"
 cache_dir="${ANI_CLI_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/ani-cli}"
-[ ! -d "$cache_dir" ] && mkdir -p "$cache_dir"
+if [ ! -d "$cache_dir" ];then
+	mkdir -p "$cache_dir"
+else
+	rm -f "$cache_dir/*"
+fi
 hist_dir="${ANI_CLI_HIST_DIR:-${XDG_STATE_HOME:-$HOME/.local/state}/ani-cli}"
 [ ! -d "$hist_dir" ] && mkdir -p "$hist_dir"
 histfile="$hist_dir/ani-hsts"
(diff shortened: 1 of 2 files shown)


ani-cli
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.4.5"
+version_number="4.4.6"
```
Derisis13, line 3: Let it be a major bump so downstream updates as well.

ani-cli
```diff
@@ -134,20 +134,24 @@ provider_init() {
...
-        # shellcheck disable=SC2059
-        ascii="$ascii$(printf "\\$oct")"
-        i=$((i + 2))
+bypass_debug() {
+    if printf "%s" "$-" | grep -q 'x'; then
+        set +x
+    elif [ "$1" -eq "0" ]; then
+        set -x
+    fi
+}
```
Derisis13, line 143: This is no longer necessarily a function as we only `set +x` in line 146 and `set -x` in 154. Minus 8 lines, and if we want to do it again we can still re-introduce the function.

ani-cli
```diff
@@ -299,7 +303,7 @@ play() {
 
 # setup
 agent="Mozilla/5.0 (Windows NT 6.1; Win64; rv:109.0) Gecko/20100101 Firefox/109.0"
-allanime_base="allanime.to"
+allanime_base="api.allanime.day"
```
Derisis13, line 306: We're looking at two constants now, one for the "frontend" and one for the "api". You don't want to replace a bunch of strings, do you...

Derisis13 (changes requested):
works well, but there are some minor adjustments I want to see for better code quality

ani-cli
```diff
@@ -180,10 +183,11 @@ get_episode_url() {
...
     episode_embed_gql="query (\$showId: String!, \$translationType: VaildTranslationTypeEnumType!, \$episodeString: String!) {    episode(        showId: \$showId        translationType: \$translationType        episodeString: \$episodeString    ) {        episodeString sourceUrls    }}"
 
-    resp=$(curl -e "https://${allanime_base}" -s --cipher "AES256-SHA256" -G "https://api.${allanime_base}/allanimeapi" --data-urlencode "variables={\"showId\":\"$id\",\"translationType\":\"$mode\",\"episodeString\":\"$ep_no\"}" --data-urlencode "query=$episode_embed_gql" -A "$agent" | tr '{}' '\n' | sed 's|\\u002F|\/|g;s|\\||g' | sed -nE 's|.*sourceUrl":"#([^"]*)".*sourceName":"([^"]*)".*|\2 :\1|p')
+    resp=$(curl -e "https://allanime.to" -s --cipher "AES256-SHA256" -G "https://${allanime_base}/api" --data-urlencode "variables={\"showId\":\"$id\",\"translationType\":\"$mode\",\"episodeString\":\"$ep_no\"}" --data-urlencode "query=$episode_embed_gql" -A "$agent" | tr '{}' '\n' | sed 's|\\u002F|\/|g;s|\\||g' | sed -nE 's|.*sourceUrl":"##([^"]*)".*sourceName":"([^"]*)".*|\2 :\1|p')
     # generate links into sequential files
     provider=1
     i=0
+    rm -f "$cache_dir/*"
```
Derisis13, line 190: this might be worth including in the initialisation

CoolnsX: okay, on it

ani-cli
CoolnsX: it's now only api.allanime.day, cause allanime.day doesn't exist

ani-cli
CoolnsX: yeah, I wrote that function cause shellcheck was complaining of && || , and elif is necessary cause If it's normal else then it would output debug value in non-debug mode
image

ani-cli
CoolnsX: on it

ani-cli
Derisis13: yes but it'd still be worth making into a variable because of how often it repeats

ani-cli
port19x: agreed

ani-cli
port19x: allanime_base is still a good idea for a variable, yes.
Especially since we already have it, I dislike the current hardcoded find-replace

ani-cli
```diff
@@ -318,7 +322,11 @@ use_external_menu="${ANI_CLI_EXTERNAL_MENU:-0}"
 [ "$use_external_menu" = "0" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-m"}"
 [ "$use_external_menu" = "1" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-multi-select"}"
 cache_dir="${ANI_CLI_CACHE_DIR:-${XDG_CACHE_HOME:-$HOME/.cache}/ani-cli}"
-[ ! -d "$cache_dir" ] && mkdir -p "$cache_dir"
+if [ ! -d "$cache_dir" ]; then
+    mkdir -p "$cache_dir"
+else
+    rm -f "$cache_dir/*"
+fi
```
port19x, line 329: No?
Cache dir should be fine to be persistent. 
Also with this you're either initializing the cache dir or removing it, which results in inconsistent behavior.
Also there doesn't need to be a `[ ! -d "$cache_dir" ]` because `mkdir -p` is already idempotent.

port19x (changes requested):
I'd prefer two argument `printf` calls for colored output. Like the printf %s content type of deal

ani-cli
CoolnsX: it's not removing cache_dir it's removing the files in it if exist

CoolnsX: image

ani-cli
CoolnsX: removed the function and replaced it with one-liners

ani-cli
port19x: still odd. If the files are truly so irrelevant to be removed after every run, why not use `mktemp`?

CoolnsX: it's because sometimes links are not generated, and it starts playing links from dangling file

ani-cli
71zenith: I've implemented the above

ani-cli
```diff
@@ -231,8 +228,8 @@ update_history() {
 download() {
     case $1 in
         *m3u8*)
-            if command -v "hls"; then
-                hls -ro "$download_dir/$2" "$1"
+            if command -v "yt-dlp" >/dev/null; then
+                yt-dlp "$1" --no-skip-unavailable-fragments --fragment-retries infinite -N 16 -o "$download_dir/$2.mp4"
```
candrapersada, line 232: `yt-dlp "$1" --no-skip-unavailable-fragments --fragment-retries infinite -N 16 --paths "$download_dir" --paths temp: ${HOME}/.cache/yt-dlp-temp --output "$2.mp4"`

candrapersada (changes requested):

candrapersada:
how to select 480p m3u8 during download
image

CoolnsX:
same ```-q 480```, doesn't matter if it's hls or mp4, works for all

ani-cli
CoolnsX: > `yt-dlp "$1" --no-skip-unavailable-fragments --fragment-retries infinite -N 16 --paths "$download_dir" --paths temp: "${HOME}/.cache/yt-dlp-temp" --output "$2.mp4"`

this is for windows right?, also you can install yt-dlp from scoop iirc

candrapersada:
> same `-q 480`, doesn't matter if it's hls or mp4, works for all

how to choose n3u8 not mp4?

ani-cli
candrapersada: what I mean is the yt-dlp temp path isn't it a matter of how to install it?

port19x:
Should be mergeable later today

ani-cli
port19x: If we explicitly set the temp path I'd be for `mktemp -d`

port19x:
Nvm, this is mergeable on the 18th at the earliest

PR #1224, opened 2023-10-07
fix: multi-selection support

AzureOrange404:
# Pull Request

## Type of change

- [x] Bug fix for [issue 1225](https://github.com/pystardust/ani-cli/issues/1225)
- [ ] Feature
- [ ] Documentation update

## Description

The check for [ $use_external_menu = 0 | 1 ] was made before the variable even gets a chance to be changed and use_external_menu was set to 0 just before that if ani-cli was started from a terminal.

BEFORE multi-selection (from the terminal with --rofi set) only worked if the env variable $ANI_CLI_EXTERNAL_MENU was set or ani-cli was not started from a terminal.

NOW multi-selection also works if the --rofi flag is set in via terminal, because the check takes place AFTER parcing command-line-arguments.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.6.1"
+version_number="4.6.2"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -309,8 +309,6 @@ esac
 
 use_external_menu="${ANI_CLI_EXTERNAL_MENU:-0}"
 [ -t 0 ] || use_external_menu=1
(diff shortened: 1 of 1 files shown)


justchokingaround (approved):

Derisis13:
it works, but I would've preferred an if-else structure

PR #1264, opened 2024-01-24
feat: no-detach

port19x:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

fixes #1261 

## Checklist

- [ ] any anime playing
- [x] bumped version
---
- [x] `-h` help info is up to date
- [x] Readme is up to date
- [x] Man page is up to date

PR #1274, opened 2024-02-04
fix: no-detach detatching

port19x:
https://github.com/pystardust/ani-cli/pull/1264#issuecomment-1925810856


port19x:
I think shellcheck is a little confused here

germaniuss:
Just tested out this MR and it does not seem to work. I feel like not redirecting the mpv output to /dev/null is necessary since when used with the kitty protocol the frames are sent through a bytestream from mpv to the terminal.

germaniuss:
> Just tested out this MR and it does not seem to work. I feel like not redirecting the mpv output to /dev/null is necessary since when used with the kitty protocol the frames are sent through a bytestream from mpv to the terminal.

By not work, I mean that the video is not showing. I no longer need to pkill the mpv process since it does not detach anymore, but the kitty protocol messages are not redirected to the terminal.

port19x:
fixed, please try again to verify

germaniuss:
Well that was fast haha, it does work now!!

port19x:
On vacation doing some random coding currently :D

Derisis13 (approved):
So far so good. Shouldn't shellcheck catch this one?

Derisis13:
oh wait, it did

port19x:
Yeah, I wanted to be smart, but shellcheck didn't let me

PR #1332, opened 2024-05-11
Add flag to allow adding an new anime to the histfile

0WN463:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

A use case that I frequently have is that I want to add an anime to my histfile so that I can watch it later.
This PR adds a flag `-a/--add` which allows us to simply add the anime without playing the episode.

For these animes, we use `unwatched` to signify the "ep_no" field in the histfile.

This flag only significant in the case of using `-c` later. 
Standard operations of `ani-cli` have no interaction with this flag.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-S` select index works
- [ ] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.8.6"
+version_number="4.9.0"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -39,6 +39,8 @@ help_info() {
     %s [options] [query] [options]
 
     Options:
+      -a, --add
+        Add an anime to the history (to watch later)
       -c, --continue
         Continue watching from history
       -d, --download
@@ -237,6 +239,16 @@ update_history() {
     mv "${histfile}.new" "$histfile"
 }
 
+add_unwatched_to_history() {
+    if grep -q -- "$id" "$histfile"; then
+        return
+    fi
+
+    cp "$histfile" "${histfile}.new"
+    printf "unwatched\t%s\t%s\n" "$id" "$title" >>"${histfile}.new"
+    mv "${histfile}.new" "$histfile"
+}
+
 download() {
     case $1 in
         *m3u8*)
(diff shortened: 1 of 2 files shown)


ani-cli
```diff
@@ -432,8 +452,7 @@ case "$search" in
         resfile="$(mktemp)"
         grep "$result" "$histfile" >"$resfile"
         read -r ep_no id title <"$resfile"
-        ep_list=$(episodes_list "$id")
-        ep_no=$(printf "%s" "$ep_list" | sed -n "/^${ep_no}$/{n;p;}") 2>/dev/null
+        process_hist_entry
```
0WN463, line 455: seems line L435 and 436 is actually duplicating `process_hist_entry`

port19x:
I kinda get it, but I don't think this feature is desirable to a broader audience

0WN463:
Why not though?

We do have the ` -D, --delete` flag, which only deletes the history, so we do consider user flows which uses the history.
The delete flag only makes sense in the history-based flow, so it means we are support this flow.

So I feel that it makes sense to bolster the history-based flow by adding this feature.

Derisis13:
This feature is not that bad, but the implementation was way too many loc. @port19x would you be against it if it were below 15 lines total?

port19x:
We most definitely support history based flow, but I believe it's unwise to let ani-cli feature-creep into a local anilist equivalent.
We can add it if derisis sees value, especially if we get it down to say +20 loc

71zenith:
+1
This is a bit too niche of a feature, i doubt anyone would be deep into this workflow.

PR #1380, opened 2024-07-16
docs: create ani-cli.cmd for windows without install ani-cli from scoop

candrapersada:


port19x (changes requested):
This, at a minimum, needs some mention in the readme.
Also I'm unsure what issue this is supposed to solve

candrapersada:
You can use `ani-cli.cmd/bat` for `ani-cli` from cmd by putting `ani-cli.cmd/bat` in the `PATH environment variable` without needing to install a `scoop`

port19x:
Not worth it

PR #1383, opened 2024-07-25
fix: steamdeck kernel detection

basicJavaCoder:
# Pull Request Template

## Type of change

- [X ] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

Ani-Cli originally checks for Steam Deck / SteamOS by checking for the default host name, which breaks the script if the host name is changed by the user. This fix checks for the kernel codename instead.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.8.10"
+version_number="4.8.11"
 
 # UI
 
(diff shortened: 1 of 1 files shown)


port19x (approved):
good catch

port19x:
[`linux-neptune`](https://github.com/neptune-mirror/linux-neptune)

PR #1386, opened 2024-07-28
fix: support using non brew IINA as a player on macOS.

asamahy:
If IINA was built from source or downloaded directly from iina.io then ani-cli wont find it unless added manually to PATH, this works aroung it without much change to rest of script.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -106,10 +106,31 @@ update_script() {
     exit 0
 }
 
+where_iina() {
+    # check if iina's symlink is present
+    if [ ! $(which iina) ]; then
+    # set the location of iina-cli to /Appliations directory
+        player_function="/Applications/IINA.app/Contents/MacOS/iina-cli"
+    fi
+    # start iina with the given arguments $1 = allanime_title, $2 = ep_no, $3 = episode
+    nohup "$player_function" --no-stdin --keep-running --mpv-force-media-title="${1}Episode ${2}" "$3"  2>&1 &
+    # set the player_function back to iina as not to break the rest of the script
+    player_function="iina"
+}
+
 # checks if dependencies are present
 dep_ch() {
     for dep; do
+    # check if iina is the player_function and if iina is installed either in PATH or in /Applications.
+    if [ "$dep " = "iina " ]; then
+        if [ $(which iina) ] || [ -f "/Applications/IINA.app/Contents/MacOS/iina-cli" ]; then
+            player_function="iina";
+        else
+            die "Program \"iina\" not found. Please install it."
+        fi
+    else
         command -v "$dep" >/dev/null || die "Program \"$dep\" not found. Please install it."
+    fi
     done
 }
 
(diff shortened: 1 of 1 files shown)


port19x (changes requested):
This is way too wordy of a change.
Please decomment and make this smaller

port19x:
Does `which iina-cli` detect your manually installed iina btw?

port19x:
Because if it does, your fix should comfortably fit into [L325](https://github.com/pystardust/ani-cli/blob/267880fc936d8c2eccfb81944ec4854a7ea79a5a/ani-cli#L325)

asamahy:
> This is way too wordy of a change. Please decomment and make this smaller

which part is wordy? the description or the title?



> Does `which iina-cli` detect your manually installed iina btw?

yeah it works for both. which_iina() only works if iina wasnt in PATH. even though it doesnt make a difference because brew already installs in `/Applications` but i didnt want to assume in case someone changed it.

> Because if it does, your fix should comfortably fit into [L325](https://github.com/pystardust/ani-cli/blob/267880fc936d8c2eccfb81944ec4854a7ea79a5a/ani-cli#L325)

it cant fit here because the rest of the script assumes the value to be "iina" in "player_function".
[L325](https://github.com/pystardust/ani-cli/blob/267880fc936d8c2eccfb81944ec4854a7ea79a5a/ani-cli#L325) is required to reach `play_episode()` [L292](https://github.com/pystardust/ani-cli/blob/267880fc936d8c2eccfb81944ec4854a7ea79a5a/ani-cli#L292);

do you want to remove the reliance on the variable `ANI_CLI_PLAYER`?

port19x:
The change is wordy, altough I see now that's a bad adjective to describe it.
I'd like to have some fallback to iina-cli if iina is not present.
Is that not possible within L325?

asamahy:
> ... I'd like to have some fallback to iina-cli if iina is not present. Is that not possible within L325?

`iina-cli` itself is not exposed to `PATH` but is symlinked to `/usr/local/bin/iina`, which is in `PATH`. Therefore, the fallback for brew iina is manually installing it in /Applications, which my change covers.

i hope i understood your point and answered adequately.

port19x:
I think I'll just need to experiment myself with the iina on my machine at home

port19x:
Alright, I can confirm that currently, out-of-band iina doesn't work and that your patch fixes that

port19x (changes requested):
let me be more specific with the changes I want.

You know the section where player_function currently gets set to iina?
There you can call a modified version of you `where_iina` function that should return the appropriate iina as a string.
`dep_ch` can probably remain unchanged, or have a very small iina catch instead of the current thing

port19x:
image

port19x:
If you can do a little testing in my follow up PR, that would be great.
I will be at home for testing in 2-3 hours at the earliest

asamahy:
srry i honestly didnt have the time to revisit this. you cant change `player_function` to anything other than `iina` without breaking playing the next episode

integrating Line 113 and 116 would remove the need to reset the `player_function` as i did there

port19x:
Good to know. Then I'll explicitly test playing the next episode

port19x:
I still included you as co-author, so you'll appear in the next release notes.
Good work for a first PR and good catch with the iina in the first place

asamahy:
thanks. new account and trying to contribute instead of just keeping them private.

PR #1398, opened 2024-08-16
fix: support using non brew iina as a player on macos

port19x:
# Pull Request Template

## Type of change

- [x] Bug fix

## Description

supersedes #1386 

## To test

- [x] the `-e` test works on mac os (could fail due to permissions)
- [x] both iinas work
- [x] failing dep check works as expected
- [x] test playing the next episode
- [x] test `-c`

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -115,7 +115,7 @@ dep_ch() {
 
 where_iina() {
     [ -n "$ANI_CLI_PLAYER" ] && echo "$ANI_CLI_PLAYER" && return 0
-    [ -n "$(command -v iina)" ] && echo "iina" && return 0
+    command -v iina >/dev/null && echo "iina" && return 0
     [ -e "/Applications/IINA.app/Contents/MacOS/iina-cli" ] && echo "/Applications/IINA.app/Contents/MacOS/iina-cli" && return 0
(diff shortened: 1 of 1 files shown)


asamahy:
`play_episode()` #L258 is dependent on the case so `player_function` need to remain `iina` and then handled inside `iina` case. otherwise it gets sent into the default.

port19x:
I see what you mean now.
*iina* should match it, will amend later

asamahy:
`next` doesnt work. after link fetching `play_episode()` doesnt seem to get the link as far as i can tell.

asamahy:
original ani-cli too doesnt have next working for me. dont know why. does it work on your side?

port19x:
I'll check when I'm home on the macbook

port19x:
I forgot :skull:, just remembered when turning on the PC at work

port19x:
This branch works flawlessly with both iina versions.
Please open another issue for your `next` issue

71zenith (changes requested):
logically looks fine, just change echo to printf for consistency
also if u could quote the $(where_iina) , cant be too safe dealing with paths

port19x:
>requests changes
>leaves

chad

PR #1402, opened 2024-08-22
feat: add -C/--copy-link to copy video link

Vosjedev:
# -C or --copy-link to copy video link

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

This feature adds the option `-C` or `--copy-link`, and it takes the link that would otherwise be passed to mpv, vlc, or a downloader, and copies it to the system clipboard. I have added support for:
- linux (detects wayland or xorg by the presence of `$WAYLAND_DISPLAY` and `$DISPLAY`, fails if both are unset)
- MacOS (using `pbcopy`)
- windows (using `C:\Windows\System32\clip.exe`)
- termux (using `termux-clipboard-set` from the `termux-api` package)

it throws an error on iOS and prints the link instead, and for everything else it defaults to linux behaviour.
I have tested this on windows, linux (arch, running hyprland), and iOS. I'm planning on testing MacOS as well, but I have to wait a bit for that.
For any future issues with this feature or issues caused by this feature you're allowed to ping/assign me to solve it (seems logical to me, if I add a feature and it breaks I have to solve it ofc :D).

Note I have not checked all things in the checklist below, and for what I didn't check I appended `(I have not tested)` to the item. These items were difficult enough for me to skip because I don't think the features would break due to this addition. `-r` seems to be broken on master already (got from AUR), so I don't know what went wrong there. I'll make a bug report if needed.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [ ] next, prev and replay work (I have not tested)
- [x] `-d` downloads work
- [ ] `-s` syncplay works (I have not tested)
- [x] `-q` quality works (probably)
- [x] `-v` vlc works
- [x] `-S` select index works
- [ ] `-r` range selection works (master seemed to be broken already?)
- [ ] `--skip` ani-skip works (I have not tested)
- [ ] `--skip-title` ani-skip title argument works (I have not tested)
- [ ] all providers return links (not necessarily on a single anime, use debug mode to confirm) (I have not tested)
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/README.md
+++ b/README.md
@@ -508,6 +508,9 @@ apk del grep sed curl fzf git aria2 ffmpeg ncurses
 - fzf - User interface
 - ani-skip (optional)
 - patch - Self updating
+- xclip - copy video link on Xorg (optional)
+- wl-clipboard - copy video link on wayland (optional)
+- termux-api - copy video link on termux (optional)
 
 ### Ani-Skip
 
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.8.11"
+version_number="4.8.12"
 
 # UI
 
@@ -41,6 +41,8 @@ help_info() {
     Options:
       -c, --continue
         Continue watching from history
+		-C, --copy-link
+		  Copy the video link to clipboard
       -d, --download
         Download the video instead of playing it
       -D, --delete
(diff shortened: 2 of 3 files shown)


port19x:
This can already be achieved today by changing the environment variable `ANI_CLI_PLAYER` to `xclip -selection c` or an equivalent wrapper.

I'm impressed with the thorough implementation and testing you have done, but caution you to raise feature requests first, or at least wait with extensive support and testing until you have the green light from the project maintainer in an open source project.

I'm declining this feature, because I believe it brings little value to the table, despite being as well documented and implemented as it is.
Don't let this discourage you from contributing to OSS, you clearly have the motivation and skill to make a positive impact

port19x:
I should also add that `ANI_CLI_PLAYER=debug` will print out the links in a pipeable way

Vosjedev:
> I'm impressed with the thorough implementation and testing you have done

Thank you!
(I was testing it on a mac while writing and found out it was broken on there, but I guess I don't have to fix a fix for it now, as I don't use a mac normally. I also found other issues with the formatting due to editor configuration issues on my side (to be precise incorrect tab usage))

> but caution you to raise feature requests first, or at least wait with extensive support and testing until you have the green light from the project maintainer in an open source project.

That might me smart, although I did plan on keeping the fork if this pr would get closed, because I have had several occasions where I would have liked to be able to do this using a quick `-C` instead of messing with environment variables.

> I'm declining this feature, because I believe it brings little value to the table, despite being as well documented and implemented as it is.

I half expected this, so no problem.

> Don't let this discourage you from contributing to OSS, you clearly have the motivation and skill to make a positive impact

Of course it won't!

Also thanks for the quick review, I had expected it to take at least a day before I would get a response

Issue #1408, opened 2024-08-31
Use flatpak mpv on other distros


randomazzy11:
**Is your feature request related to a problem? Please describe.**

I use fedora silverblue, which is immutable like steamos
and i used the install script for steamos to install ani-cli
but then mpv didnt work, and i quickly looked at the code and saw
that flatpak mpv is only used when steamos is detected.

**Describe the solution you'd like**

Maybe check first for mpv in the regular path first
and then maybe if its not there check for
flatpak.

**Describe alternatives you've considered**

Add a flatpak flag (since i didnt see any) to force
using flatpak mpv.
I also considered modifying a line in the code by changing
the normal mpv under "linux" to be flatpak mpv but i realised
that an update might wipe my changes.
I also would be able to just download an episode
and play it that way but thats not really elegant

**Additional context**

I could install the ani-cli package using rpm-ostree (which is like dnf but for silverblue) 
but i dont really like using rpm-ostree and i only install apps using it if i really have to

Derisis13:
It surprises me that we don't have an argument for flatpak, I'll implement it once I get home. 
While I do that, use the env var `ANI_CLI_PLAYER` as a workaround.
`env ANI_CLI_PLAYER="flatpak_mpv" ani-cli`

randomazzy11:
Thanks :) some people need that flatpak argument

randomazzy11:
Thanks :) some people need that flatpak argument

PR #1411, opened 2024-09-05
feat: added flag for mpv as flatpak

Derisis13:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

Comply with request #1408

## Checklist

- [x] any anime playing
- [ ] bumped version
---
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-S` select index works
- [ ] `--skip` ani-skip works
---
- [ ] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.9.0"
+version_number="4.9.1"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -45,6 +45,8 @@ help_info() {
         Download the video instead of playing it
       -D, --delete
         Delete history
+      -f, --flatpak-mpv
+        Use mpv (flatpak) to play the video
       -l, --logview
         Show logs
       -s, --syncplay
@@ -364,6 +366,8 @@ while [ $# -gt 0 ]; do
                 *) player_function="vlc" ;;
             esac
             ;;
+        -f | --flatpak-mpv)
+            player_function="flatpak_mpv" ;;
         -s | --syncplay)
             case "$(uname -s)" in
                 Darwin*) player_function="/Applications/Syncplay.app/Contents/MacOS/syncplay" ;;
--- a/ani-cli
+++ b/ani-cli
@@ -45,8 +45,6 @@ help_info() {
         Download the video instead of playing it
       -D, --delete
         Delete history
-      -f, --flatpak-mpv
-        Use mpv (flatpak) to play the video
       -l, --logview
         Show logs
       -s, --syncplay
@@ -122,6 +120,12 @@ where_iina() {
     command -v iina >/dev/null && printf "iina" && return 0
     [ -e "/Applications/IINA.app/Contents/MacOS/iina-cli" ] && echo "/Applications/IINA.app/Contents/MacOS/iina-cli" && return 0
     dep_ch iina # exit with formating
+}
+
+where_mpv() {
+    command -v "$ANI_CLI_PLAYER" >/dev/null && printf "%s" "$ANI_CLI_PLAYER" && return 0
+    command -v "flatpak" >/dev/null && flatpak info io.mpv.Mpv >/dev/null 2>&1 && printf "%s" "flatpak_mpv" && return 0
+    dep_ch mpv # exit with formating
 }
 
 # SCRAPING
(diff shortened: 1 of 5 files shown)


port19x:
Didn't we already have that at some point?

Derisis13:
> Didn't we already have that at some point?

We may have. But it seems like there's a need for this. I see no reason why we shouldn't have it.

port19x:
Because we already have a flatpak_mpv player for steamdeck implemented.
I'd rather have [player detection](https://github.com/pystardust/ani-cli/blob/master/ani-cli#L335-L342) fall back to flatpack mpv when a regular mpv can't be detected.
A similar pattern to what I want is already used for iina, where manually installed iina needs different handling than iina installed via homebrew

port19x:
This would have the additional upside of consolidating case branches with steamdeck

Derisis13:
Ok, can do. It makes sense not to not trust it onto our users to know which package format they use

ani-cli
```diff
@@ -337,10 +341,12 @@ quality="${ANI_CLI_QUALITY:-best}"
 case "$(uname -a)" in
     *Darwin*) player_function="$(where_iina)" ;;                      # mac OS
     *ndroid*) player_function="${ANI_CLI_PLAYER:-android_mpv}" ;;     # Android OS (termux)
-    *neptune*) player_function="${ANI_CLI_PLAYER:-flatpak_mpv}" ;;    # steamdeck OS
     *MINGW* | *WSL2*) player_function="${ANI_CLI_PLAYER:-mpv.exe}" ;; # Windows OS
     *ish*) player_function="${ANI_CLI_PLAYER:-iSH}" ;;                # iOS (iSH)
-    *) player_function="${ANI_CLI_PLAYER:-mpv}" ;;                    # Linux OS
+    *) 
+      player_function="${ANI_CLI_PLAYER:-$(where_mpv)}"               # Linux OS
+      [ $? -eq 0 ] || exit $?
```
port19x, line 348: This part is actually not necessary, since `dep_ch` will already exit

port19x (changes requested):

Derisis13:
I've tested what I could on linux, but @ykhan21 I'd like you to test it on windows if it correctly detects installed and uninstalled mpv, and @Nannk to check it on the steamdeck if it still finds the flatpak version.
Also @port19x I think there's a little issue with `where_iina()` which I couldn't test: when the subshell exits, the main program continues uninterrupted. I suggest to use a similar structure like I did with `where_mpv()`

ani-cli
Derisis13: The whole reason I did that so we can detect if either packaging format has mpv. I'm open to suggestions for better options.

port19x:
iirc it worked fine in my testing. `return` exits subshells, `exit` always aborts the shell script.

[SO Post](https://stackoverflow.com/questions/4419952/difference-between-return-and-exit-in-bash-functions)

Where is this not the case?

port19x:
Could it be because you use a default assignment? `${ANI_CLI_PLAYER:-$(where_mpv)}`

In where iina, that is checked within the function, but I'd be surprised if that changes things

Derisis13:
> iirc it worked fine in my testing on mac os. `return` exits subshells, `exit` always aborts the shell script.
> 
> [SO Post](https://stackoverflow.com/questions/4419952/difference-between-return-and-exit-in-bash-functions)
> 
> Where is this not the case?

Linux and dash. Same if the default assignment is not present, like with iina.

Derisis13:
> iirc it worked fine in my testing on mac os. `return` exits subshells, `exit` always aborts the shell script.
> 
> [SO Post](https://stackoverflow.com/questions/4419952/difference-between-return-and-exit-in-bash-functions)
> 
> Where is this not the case?

Linux and dash. Same if the default assignment is not present, like with iina.

And checking the [shell specification's related section](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#exit) reveals that `exit` only returns from the current subshell and continues from the environment it was called from.

port19x:
Alright, I'm taking the L

Derisis13:
Now then why haven't the two remaining checks ran?

port19x:
> Now then why haven't the two remaining checks ran?

I have no idea. I added the completion files to path ignore, but you haven't changed those in this PR, so I'm surprised they aren't running. Github Actions is a bit fragile at times.

Wanna patch my where_iina to be more in line with your where_mpv?

Let's see if github wakes from its slumber after another commit

Derisis13:
> Wanna patch my where_iina to be more in line with your where_mpv?

Sure, let's see

port19x:
Now that's a surprise. Perhaps a brown out on githubs part to bully people into migrating away from a deprecated nodejs version (used by checkout v3)

port19x (approved):

PR #1419, opened 2024-09-12
feat: tab-completion for zsh and bash

lucathehun:
This completion option is for zsh **only** at the moment, if you like what you see or have suggestions for what other shells I should make it for tell me and I'll do my best.

```zsh
ani-cli --completion > _ani-cli && echo 'source /path/to/_ani-cli' >> ~/.zshrc
```

Really enjoying the tool itself btw 💜 

image


port19x:
Bash completion would be nice

lucathehun:
> Bash completion would be nice

I'll work on that tomorrow, if I got time.

port19x:
Also, I'd prefer to have the completion data separate file that people can install as you show in your screenshot

port19x:
Otherwise this looks good

lucathehun:
The idea was tho add options for every shell that is gonna be implemented for it. 
So that u can just type

```
ani-cli --completion <shell> 
```

Didn't add the last part since I just tried it with zsh for a start.

lucathehun:
Got the bash completion done, u just have to type,
```
ani-cli --completion
```
and it's gonna detect the shell u are using and put out the zsh-/bash-completion script.
Otherwise it tells u that it's not supported. 

It worked perfectly on my system, if you got any problems with it, just tell me.

Thank you!

lucathehun:
Now it should work for bash and zsh.

```
ani-cli --completion > _ani-cli && source _ani-cli
```

detects the shell automatically

port19x:
The problem I have with this is that the script gets bloated 120 lines for a glorified print.

Please make these things separate files in the repo and document their use in the README

lucathehun:
I have 2 options for you here. 

One where you just need to execute the script and another one that requires you to move and source the completion file one your own. 

Which should it be?

port19x:
Thanks, I prefer the manual way.
Is the new line at the top intentional for those files btw?

lucathehun:
Okay then I'll edit it according to that.

No I copied them from the original script and it just happens to be the case that nvim pastes under and not on the line you are on. 

And one more thing... how are the people that use their package manager to install ani-cli use this now? 
That was the original thought behind me adding it to the main script.

lucathehun:
I mean without doing an extra step and downloading from github anyway

port19x:
Packagers could work around this, or they would need to fetch the completion info manually.

I believe this is fine, because the people that would see `--completion` in help output and successfully set it up that way are the same people that will carefully read the readme anyway

lucathehun:
Hope that it's to your liking now @port19x !
And thanks for being patient with me. 💜

port19x (approved):

port19x:
Thanks for the contribution

PR #1426, opened 2024-09-18
ci: shellcheck completion file annoyance

port19x:
cow-cow


port19x:
Today is not my brightest. This ought to fix things

71zenith (approved):

Issue #1429, opened 2024-09-22
Receive "Episode not released!" even though the episode is released on allanime.to


Lugas7:
Version: 4.9.0
OS: Arch Linux
Shell: bash
Anime: Kami No Tou (Tower of god season 2)

** description **

Get "Episode not released!" when the episode is realeased on allanime.to.

**Steps To Reproduce**

Happens when I try to watch an episode which is tracked in ani-cli and can thence be selected, but it fails. Think it has something to do with the episode only recently being released (as of my attempt, the episode has only been released 5 hours ago on allanime.to)

**Expected behavior**
Should be able to watch episode

HirschBerge:
Also happening on 4.9.0 on NixOS unstable. Has been happening with every episode released in the last 24-36 hours or so that I'm watching. Tower of God, Monogatari, Fairy Tail 100YQ, etc.
Saw something similar a few weeks ago that eventually resolved itself I believe.

71zenith:
We scrape only a small subset of video providers from allanime. Most of the times when a new episode is released it only has providers we dont scrape.

Lugas7:
Ah ok, is there an issue for increasing the number of providers scraped? I might be interested in attempting developing that at some point when I have time

71zenith:
> Ah ok, is there an issue for increasing the number of providers scraped? I might be interested in attempting developing that at some point when I have time

No mostly, as long as the code isnt too complicated. 
you can start by looking at the current scrapers https://github.com/pystardust/ani-cli/blob/5daa876a3c88544ee7a21aac4704e14be519c25a/ani-cli#L131-L197

71zenith:
Generally we don't add more cause the current providers are the most stable and don't break unexpectedly

Lugas7:
One quick improvement would be to change the error message in this event, it is in fact incorrect to say that the episode is not released yet. Perhaps something along the lines of: "episode has not been released in a provider we support" may be better.

I would not have made this issue as a bug if I got a better error message.

Issue #1391, opened 2024-08-05
Wrong anime title due to wrong extraction from regex caused by escaped `"`.


0WN463:
## **Make sure you are on the latest version by running `ani-cli -U`**

**Metadata (please complete the following information)**
Version: 4.8.11
OS: Arch Linux
Shell: zsh
Anime: [e.g. flcl] (if applicable) Oshi no Ko Season 2

**Describe the bug**
Oshi no Ko Season 2's title is bugged.

It says
```
1 \ (5 Episodes)
```

**Steps To Reproduce**

1. Run `ani-cli Oshi`
2. First episode on the list is bugged

**Expected behavior**
Title should appear normally

**Additional context**
Looking through the script, it seems like the JSON is
```
          "name": "\"Oshi no Ko\" Season 2",
```

And the escaped `"` is being treated wrongly and thus the wrong title is being extracted

port19x:
Reproduced

<details>
<pre>
+ version_number=4.8.11
+ agent='Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:109.0) Gecko/20100101 Firefox/121.0'
+ allanime_refr=https://allanime.to
+ allanime_base=allanime.day
+ allanime_api=https://api.allanime.day
+ mode=sub
+ download_dir=.
+ quality=best
+ case "$(uname -a)" in
++ uname -a
+ player_function=iina
+ no_detach=0
+ exit_after_play=0
+ use_external_menu=0
+ skip_intro=0
+ skip_title=
+ '[' -t 0 ']'
+ hist_dir=/Users/port19/.local/state/ani-cli
+ '[' '!' -d /Users/port19/.local/state/ani-cli ']'
+ histfile=/Users/port19/.local/state/ani-cli/ani-hsts
+ '[' '!' -f /Users/port19/.local/state/ani-cli/ani-hsts ']'
+ search=scrape
+ '[' 1 -gt 0 ']'
+ case "$1" in
++ printf %s ' Oshi no Ko'
++ sed 's|^ ||;s| |+|g'
+ query=Oshi+no+Ko
+ shift
+ '[' 0 -gt 0 ']'
+ '[' 0 = 0 ']'
+ multi_selection_flag=-m
+ '[' 0 = 1 ']'
+ printf '\33[2K\r\033[1;34mChecking dependencies...\033[0m\n'
+ dep_ch curl sed grep
+ for dep in '"$@"'
+ command -v curl
+ for dep in '"$@"'
[...]

PR #1434, opened 2024-10-06
fix: Error logic and updated message and mis-escaped quotes

HirschBerge:
Checks to see if the episode selected is in the list of scraped episodes available for that show and season. Updated error message to reflect result

# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

Adds additional error messages as discussed in #1429. Previously, any time there was an error, it would always say "Episode not released!" which is not true sometimes. This happens in the case of an episode being released, but not pulling a source correctly (or source not available). This just adds a simple check to see if the $ep_no is in $ep_list. If it is and $episode is empty, it states that the episode is released, but no source available, else, it falls back to the current error message.

While I was here, I implemented 71zenith's suggestion in #1391 to fix issues with specifically Oshi No Ko's second season. It didn't quite work as perfectly as I hoped, so I had to add an extra pipe to remove the literal \"'s, but it is tested and working with the above use case.

Sorry for making two unrelated fixes in one PR, I didn't see any guidelines against it though. Please let me know if there are better, or more polished ways to implement my fixes. I'm not as strong in bash as I am some other langs.
Thanks!

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -205,7 +205,7 @@ search_anime() {
     search_gql="query(        \$search: SearchInput        \$limit: Int        \$page: Int        \$translationType: VaildTranslationTypeEnumType        \$countryOrigin: VaildCountryOriginEnumType    ) {    shows(        search: \$search        limit: \$limit        page: \$page        translationType: \$translationType        countryOrigin: \$countryOrigin    ) {        edges {            _id name availableEpisodes __typename       }    }}"
 
     curl -e "$allanime_refr" -s -G "${allanime_api}/api" --data-urlencode "variables={\"search\":{\"allowAdult\":false,\"allowUnknown\":false,\"query\":\"$1\"},\"limit\":40,\"page\":1,\"translationType\":\"$mode\",\"countryOrigin\":\"ALL\"}" --data-urlencode "query=$search_gql" -A "$agent" | sed 's|Show|\
-|g' | sed -nE "s|.*_id\":\"([^\"]*)\",\"name\":\"([^\"]*)\".*${mode}\":([1-9][^,]*).*|\1	\2 (\3 episodes)|p"
+| g' | sed -nE "s|.*_id\":\"([^\"]*)\",\"name\":\"(.+)\",.*${mode}\":([1-9][^,]*).*|\1\t\2 (\3 episodes)|p" | sed 's/\\"//g'
 }
 
 time_until_next_ep() {
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.9.3"
+version_number="4.9.4"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.9.4" 
+version_number="4.9.4"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -193,7 +193,11 @@ get_episode_url() {
     links=$(cat "$cache_dir"/* | sed 's|^Mp4-||g;/http/!d;/Alt/d' | sort -g -r -s)
     rm -r "$cache_dir"
     episode=$(select_quality "$quality")
-    [ -z "$episode" ] && die "Episode not released!"
+    if echo "$ep_list" | grep -q "^$ep_no$"; then
+        [ -z "$episode" ] && die "Episode is released, but no valid sources!"
+    else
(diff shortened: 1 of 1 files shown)


ani-cli
```diff
@@ -193,15 +193,19 @@ get_episode_url() {
     links=$(cat "$cache_dir"/* | sed 's|^Mp4-||g;/http/!d;/Alt/d' | sort -g -r -s)
     rm -r "$cache_dir"
     episode=$(select_quality "$quality")
-    [ -z "$episode" ] && die "Episode not released!"
+    if echo "$ep_list" | grep -q "^$ep_no$"; then
```
Derisis13, line 196: Please use printf, it's the standard for the script. echo's behavior is implementations-specific and not standardized in POSIX

Derisis13 (changes requested):
Looks promising, although I have not tried it. My only issue is so far the echo statement.

ani-cli
HirschBerge: Thanks for the feedback! I have changed to using printf. I am not very familiar with what all is POSIX compliant and just assumed something so standard would be. This is tested working in my same test cases, so I think we're good on that front.

Note: in the iina function there is an echo as well. It's mac-specific, so it may not matter, and I don't have one to test with. I just caught it when searching for echo in neovim.

ani-cli
```diff
@@ -193,15 +193,19 @@ get_episode_url() {
...
+    fi
 }
 
 # search the query and give results
 search_anime() {
     search_gql="query(        \$search: SearchInput        \$limit: Int        \$page: Int        \$translationType: VaildTranslationTypeEnumType        \$countryOrigin: VaildCountryOriginEnumType    ) {    shows(        search: \$search        limit: \$limit        page: \$page        translationType: \$translationType        countryOrigin: \$countryOrigin    ) {        edges {            _id name availableEpisodes __typename       }    }}"
 
     curl -e "$allanime_refr" -s -G "${allanime_api}/api" --data-urlencode "variables={\"search\":{\"allowAdult\":false,\"allowUnknown\":false,\"query\":\"$1\"},\"limit\":40,\"page\":1,\"translationType\":\"$mode\",\"countryOrigin\":\"ALL\"}" --data-urlencode "query=$search_gql" -A "$agent" | sed 's|Show|\
-|g' | sed -nE "s|.*_id\":\"([^\"]*)\",\"name\":\"([^\"]*)\".*${mode}\":([1-9][^,]*).*|\1	\2 (\3 episodes)|p"
+| g' | sed -nE "s|.*_id\":\"([^\"]*)\",\"name\":\"(.+)\",.*${mode}\":([1-9][^,]*).*|\1\t\2 (\3 episodes)|p" | sed 's/\\"//g'
```
port19x, line 208: The literal tab may be necessary on bsd systems, those have trouble with whitespace regex characters, which is why we have some ugly literal newlines in a few places

HirschBerge: I have a commit ready to go to fix that. I searched for other \t-s, and see quite a few. Most are within `printf` statements, which it looks like supports the \t. I did see one in a `tr` (25) and a few within another `sed` (244). I also see \t mentioned in the `sed` docs for bsd, but I didn't quite get the context.
I don't know if those are worth getting into though in this PR.
I can just commit the literal tab for this particular instance and call it there.

port19x: Sed on 244 is fishy too then, but the other coreutils should be fine.
It was pretty much only sed that gave us issues

HirschBerge: I don't want to overstep, but i can push a commit removing the ones on 244.  Pretty quick fix.

port19x: Sorry for the late reply. Yes, please do. Otherwise this PR is about ready

HirschBerge: No worries! Pushed that commit and I believe it's RTG!

port19x (approved):

port19x:
Note I have not done any testing on this PR

port19x:
@Derisis13 You need to revise your review for this to become mergeable

Derisis13 (approved):
Sorry, go ahead

HirschBerge:
Thank you both!

PR #1440, opened 2024-10-13
fix: remove hostname from uname output in OS detection

VAISHAKH-GK:
# fix: remove hostname from uname output in OS detection

## Type of change

- [X] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

Using $(uname -a | awk '{$2=""; print $0}') instead of just uname -a helps remove the hostname from the output. since hostname can vary between systems it may lead to incorrect OS detection. By stripping out the hostname case statement accurately matches the relevant OS identifiers.

In my system script not work properly because my hostname is "vaishakh" , because it contains "ish" the case identify my linux device as ios device and ani-cli failed to play the video.

## Checklist

- [X] any anime playing
- [X] bumped version
---
- [X] `-d` downloads work
- [X] `-s` syncplay works
- [X] `-q` quality works
- [X] `-v` vlc works
- [X] `-e` select episode works
- [X] `-S` select index works
- [X] `-r` range selection works
- [X] `--skip` ani-skip works
- [X] `--skip-title` ani-skip title argument works
- [X] `--no-detach` no detach works
- [X] `--dub` and regular (sub) mode both work
---
- [X] Readme is up to date
- [X] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -335,7 +335,7 @@ mode="${ANI_CLI_MODE:-sub}"
 download_dir="${ANI_CLI_DOWNLOAD_DIR:-.}"
 log_episode="${ANI_CLI_LOG:-1}"
 quality="${ANI_CLI_QUALITY:-best}"
-case "$(uname -a)" in
+case "$(uname -a | cut -d " " -f 1,3-)" in
     *Darwin*) player_function="${ANI_CLI_PLAYER:-$(where_iina)}" ;;   # mac OS
     *ndroid*) player_function="${ANI_CLI_PLAYER:-android_mpv}" ;;     # Android OS (termux)
     *MINGW* | *WSL2*) player_function="${ANI_CLI_PLAYER:-mpv.exe}" ;; # Windows OS
@@ -359,7 +359,7 @@ search="${ANI_CLI_DEFAULT_SOURCE:-scrape}"
 while [ $# -gt 0 ]; do
     case "$1" in
         -v | --vlc)
-            case "$(uname -a)" in
+            case "$(uname -a | cut -d " " -f 1,3-)" in
                 *ndroid*) player_function="android_vlc" ;;
                 MINGW* | *WSL2*) player_function="vlc.exe" ;;
                 *ish*) player_function="iSH" ;;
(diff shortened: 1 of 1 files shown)


HirschBerge:
```
  check-no-awk:
    name: No Awk Allowed
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v4
    - name: verify that noone added awk
      run: '! grep awk "./ani-cli"'
   ```
awk is frowned upon. I would probably suggest `cut` something like `uname -a | cut -d ' ' -f 1,3-` at first blush does the job better even than awk since it doesn't leave two spaces in the output

**Comment by port19x** [comment]
@HirschBerge is exactly right. We have a no awk policy over here

File: ani-cli [hunk 1]
```diff
@@ -335,7 +335,7 @@ mode="${ANI_CLI_MODE:-sub}"
 download_dir="${ANI_CLI_DOWNLOAD_DIR:-.}"
 log_episode="${ANI_CLI_LOG:-1}"
 quality="${ANI_CLI_QUALITY:-best}"
-case "$(uname -a)" in
+case "$(uname -a | awk '{$2=""; print $0}')" in
```
port19x, line 338:
```suggestion
case "$(uname -a | cut -d " " -f 1,3-)" in
```

ani-cli
```diff
@@ -359,7 +359,7 @@ search="${ANI_CLI_DEFAULT_SOURCE:-scrape}"
 while [ $# -gt 0 ]; do
     case "$1" in
         -v | --vlc)
-            case "$(uname -a)" in
+            case "$(uname -a | awk '{$2=""; print $0}')" in
```
port19x, line 362:
```suggestion
            case "$(uname -a | cut -d " " -f 1,3-)" in
```

port19x (changes requested):
See my suggestions for how to get rid of the awk dependency.

Your change itself is well reasoned and a good contribution

VAISHAKH-GK:
Thanks for the suggestion i will remove awk and will update the pull request

port19x (approved):
thx for the contrib :crown:

VAISHAKH-GK:
> thx for the contrib 👑

Happy to be helpful and thanks for the understanding

PR #1445, opened 2024-10-18
fix:avoids reencoding the + while running time_until_next_ep

HirschBerge:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

Hey, it's me again. ha. I hope this one is simpler and a bit more well tested.

I noticed that checking for the next episode countdown only works when you use `-N` or `--nextep-countdown` by itself and did not provide a search directly in the cli command.
I narrowed down the reasoning to two things:
This snippet automatically replaces spaces with `+`
https://github.com/pystardust/ani-cli/blob/f2e81120c076d5e56e2420a08e3310fc61b4c0f7/ani-cli#L435
and the `curl` in the function re-encodes the `+` to become `%2B` when I look the verbose curl output.
https://github.com/pystardust/ani-cli/blob/f2e81120c076d5e56e2420a08e3310fc61b4c0f7/ani-cli#L221
The only reason that it works at all when using the built-in query picker is because the function is called one line before that query gets its spaces converted to `+`'s. I suspect this was known about when the feature was added due to the function call placement, but the author didn't realize it didn't work for queries from the cli

https://github.com/pystardust/ani-cli/blob/f2e81120c076d5e56e2420a08e3310fc61b4c0f7/ani-cli#L479-L481

I do not know a lot about curl, so Idk if there's an easier way to avoid re-encoding the `+` during the call, I tried just removing the flag, but that broke harder, so I just decided to use sed to return the spaces.

Note, I didn't test any anime that have a literal + in the title, I don't know if that exists even, so there's that.

## Checklist

- [x] any anime playing
- [ ] bumped version
---
I don't really want to go through and click all of these, i only changed the -N function so it'll all work
- [x] `--nextep-countdown`works
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-S` select index works
- [ ] `--skip` ani-skip works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -218,7 +218,8 @@ search_anime() {
 
 time_until_next_ep() {
     animeschedule="https://animeschedule.net"
-    curl -s -G "$animeschedule/api/v3/anime" --data-urlencode "q=$1" | sed 's|"id"|\n|g' | sed -nE 's|.*,"route":"([^"]*)","premier.*|\1|p' | while read -r anime; do
+    avoids_reencoding="$(printf "%s" "$1" |sed 's/+/ /g')"
+    curl -s -G "$animeschedule/api/v3/anime" --data-urlencode "q=$avoids_reencoding" | sed 's|"id"|\n|g' | sed -nE 's|.*,"route":"([^"]*)","premier.*|\1|p' | while read -r anime; do
         data=$(curl -s "$animeschedule/anime/$anime" | sed '1,/"anime-header-list-buttons-wrapper"/d' | sed -nE 's|.*countdown-time-raw" datetime="([^"]*)">.*|Next Raw Release: \1|p;s|.*countdown-time" datetime="([^"]*)">.*|Next Sub Release: \1|p;s|.*english-title">([^<]*)<.*|English Title: \1|p;s|.*main-title".*>([^<]*)<.*|Japanese Title: \1|p')
         status="Ongoing"
         color="33"


port19x:
Just seeing this now. It's on my radar, but it might take a couple more days until I take some time to fully review the issue and make sure there is no related problems

CoolnsX:
I think passing the query directly in url instead of --data-urlencode should work

CoolnsX:
or just replace --data-urlencode to --data-raw in that particular curl will also work

CoolnsX:
and you can remove that `avoids_reencoding="$(printf "%s" "$1" | sed 's/+/ /g')" ` reassignment and it will be only 1 line change

just do --
- either
```
     curl -s -G "$animeschedule/api/v3/anime" --data-raw "q=$1"
```
- or
```
     curl -s -G "$animeschedule/api/v3/anime?q=$1"
```

both works

HirschBerge:
Nice! Good suggestion, though waking up to like 8 notifications did make me concerned, lol

CoolnsX:
LGTM..

HirschBerge:
Is working on my end FWIW

port19x (approved):
thanks for the contribution and patience

PR #1479, opened 2025-01-29
feat: fetch chapter information regardless of --skip argument

topazrn:
## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

I'd like to have chapter information even without the --skip argument. This way, depending on my mood, I can manually skip or re-watch the op or ed by moving between chapters.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -275,8 +275,7 @@ download() {
 play_episode() {
     [ "$log_episode" = 1 ] && [ "$player_function" != "debug" ] && [ "$player_function" != "download" ] && command -v logger >/dev/null && logger -t ani-cli "${allanime_title}${ep_no}"
     skip_flag="$(ani-skip -q "$mal_id" -e "$ep_no")"
-    skip_flag_arr=($skip_flag)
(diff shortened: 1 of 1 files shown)


Derisis13:
Why would we do this? Can you not use the --skip argument?

Derisis13:
Also the fact we have --no-detach not working might be worthvlooking into

topazrn:
I think the --skip argument should only be used for when we want to skip the op and ed. Having a chapter information is nice whether we want to skip the op and ed or not.

Turns out the --no-detach has always worked. I just misinterpreted what it was supposed to do.

port19x:
How does this behave when ani-skip is not installed?

PR #1484, opened 2025-02-11
ani-cli : Improve sub/dub availability checks and user prompts 

vishalk17:
## Type of Change

- Feature

### Key Changes:

- **Audio Selection Prompt:**  
  After selecting an anime, the script checks if subbed and/or dubbed episodes are available.

- **User Choice:**  
  If both versions are available, you’ll be prompted to choose between them.  

- **Automatic Sub Selection:**  
  If only the sub (subtitled) version is available and no dub (dubbed) version exists:  
  - The script will automatically set the mode to "sub" without prompting the user to select the audio type.  
  - This ensures seamless playback without unnecessary prompts when only one audio type is available.

- **Backward Compatibility:**  
  Retains the `--dub` flag to force the playback of dubbed episodes without prompting.  
  - If the `--dub` flag is provided, the script will attempt to play dubbed episodes directly.  
  - If dubbed episodes are not available, an appropriate error message will be displayed.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -462,6 +462,32 @@ case "$search" in
         [ -z "${index##*[!0-9]*}" ] || id=$(printf "%s" "$anime_list" | sed -n "${index}p" | cut -f1)
         [ -z "$id" ] && exit 1
         title=$(printf "%s" "$anime_list" | grep "$id" | cut -f2 | sed 's/ - episode.*//')
+        # Check sub/dub availability and prompt
+        original_mode="$mode"
+        mode="sub"
+        sub_ep_list=$(episodes_list "$id")
+        mode="dub"
+        dub_ep_list=$(episodes_list "$id")
+        mode="$original_mode"
+        sub_available=$([ -n "$sub_ep_list" ] && echo 1 || echo 0)
+        dub_available=$([ -n "$dub_ep_list" ] && echo 1 || echo 0)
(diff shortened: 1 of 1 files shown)


vishalk17:
> I'd guess 80-90% of our users use subs exclusively. Having a prompt for every anime with dubs available would be straight up obnoxious to these people.
> 
> What about falling back to subs if `--dub` is specified but a dub is not available?

@Derisis13 
-------------------
I believe most people prefer dubbed versions if available. If not, they likely prefer subs.

There might be two cases:

    Watching Old Anime (Both Sub/Dub Available):
    If the dubbed version is already there, why would I care to watch it with subs?

    Watching New Anime (Only Sub Available):
    In this case, the choice would naturally be subs because, as an anime fan, I wouldn’t want to wait for the dubbed version to be released to watch a masterpiece.
 ----------------------
 Regarding the second question:

The --dub flag clearly indicates that the user wants to watch the dubbed version. If I build the logic as you suggested (falling back to subs when a dub isn’t available), it defeats the purpose of providing that flag.

However, in such cases, we could prompt the user:
"Dubbed version not available. Subbed version is available. Would you like to proceed with subs? (Y/N)"

This maintains the meaning of the --dub flag while still offering flexibility when dubs are unavailable.
------------------------
Agree?

mrfluffy-dev:
Ohh hell nah

justchokingaround:
no

PR #1548, opened 2025-07-15
fix: log debug outputs to stderr

sudoAlphaX:
# Pull Request Template

## Description

Send informations like "Checking dependencies", "Playing Episode n", etc
to stderr so it can be routed easily to /dev/null if needed.

Clear screen using tput clear only if player_function is not dump or
download so debug/download statuses are preserved.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -357,9 +357,9 @@ play() {
         range=$(printf "%s\n" "$ep_list" | sed -nE "/^${start}\$/,/^${end}\$/p")
         [ -z "$range" ] && die "Invalid range!"
         for i in $range; do
-            tput clear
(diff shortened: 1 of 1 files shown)


Derisis13:
I like the tput part, not sure about the stderr... Is it common to print info messages to stderr?
Also, may we know what the main purpose of this change is?

sudoAlphaX:
The main purpose of this patch is:
1. I wanted to view the errors of previous downloads
2. I wanted to dump the direct download links so i can download them later/using another tools.

port19x:
I mean I guess you don't really violate any rules by writing to stderr, but I don't see a point either, as for debugging we advise `sh -x`
https://pubs.opengroup.org/onlinepubs/9699919799/

Closing this because I don't see any point

PR #1607, opened 2026-02-24
fix: improve downloads (prefer yt-dlp), flexible deps, and update msg

makise-ui:
# Pull Request Template

## Type of change

- [x] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

This PR improves the robustness of the download functionality by preferring `yt-dlp` for both m3u8 and direct mp4 links. It also relaxes the dependency checks so that [ani-cli](cci:7://file:///home/kurisu/ani-cli/ani-cli:0:0-0:0) doesn't fail if `aria2c` is missing but `yt-dlp` is present. Finally, it improves the error message when self-update fails.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-e` (select episode aka -r range selection) works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -299,7 +299,7 @@ download() {
             # [ -e "$download_dir/$2.vtt" ] && ffmpeg -i "$download_dir/$2.mp4" -i "$download_dir/$2.vtt" -c copy -c:s mov_text "$download_dir/$2.bak.mp4" && mv "$download_dir/$2.bak.mp4" "$download_dir/$2.mp4"
             ;;
         *)
-            command -v "yt-dlp" >/dev/null && t-dlp --referer "$allanime_refr" -o "$download_dir/$2.mp4" "$1"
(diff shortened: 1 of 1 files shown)


makise-ui:
The main reason is that yt-dlp is already used in the script for m3u8 downloads, so it makes sense to also use it for direct mp4 links. This way users only need yt-dlp installed instead of both yt-dlp and aria2c. It also has better retry/error handling built in. aria2c is still kept as a fallback nothing breaks if someone doesn't have yt-dlp

port19x (changes requested):
x

PR #1622, opened 2026-03-25
docs: add Sho to homies (requesting a shoutout)

scapy47:
## Type of change

- [ ] Bug fix
- [ ] Feature
- [x] Documentation update

## Description

hi i am creater of a project inspired from ani-cli [Shio](https://github.com/Scapy47/shio). can i get shoutout in project README.


Derisis13:
I suggest to work on your project a little more, and once you feel like it's stable, tried and tested and nearing its final form, then let's discuss this again. As of now it's too young and I don't want to add it to then have to remove in two months.

PR #1650, opened 2026-04-16
fix: allanime openssl aes-256-ctr decryption

### LINKED ISSUE #1647: ani-cli suddenly stopped working.
Version: 
`❯ ani-cli -V
4.11.0`
OS: Linux 6.19.12-1-cachyos
Shell: bash
Anime: All

**Describe the bug**
```
❯ ani-cli -c
Specified quality not found, defaulting to best
Episode is released, but no valid sources!
 
❯ ani-cli kumo desu ka nani ka
Specified quality not found, defaulting to best
Episode is released, but no valid sources!
````

**Steps To Reproduce**

1. Run `ani-cli -c`
2. Choose any anime

1. Run 'ani-cli whatever-anime'
2. Choose any episode

**Expected behavior**
Should get the episode into MPV player

**Additional context**

It stopped working today, as soon as I woke up.

### PR DESCRIPTION
**Pull request by justchokingaround**
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

*ramble here*

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -212,7 +212,7 @@ decode_tobeparsed() {
     blob="$1"
     tmp="$(mktemp)"
     ct="$(mktemp)"
-    printf '%s' "$blob" | openssl enc -d -base64 -A >"$tmp"
+    printf '%s' "$blob" | base64 -d >"$tmp"
     len="$(wc -c <"$tmp" | tr -d ' ')"
     iv="$(dd if="$tmp" bs=1 count=12 2>/dev/null | od -A n -t x1 | tr -d ' \n')"
     ct_len=$((len - 28))
--- a/ani-cli
+++ b/ani-cli
@@ -506,7 +506,11 @@ done
 [ "$use_external_menu" = "1" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-multi-select"}"
 [ "$external_menu_normal_window" = "1" ] && external_menu_args="-normal-window"
 printf "\33[2K\r\033[1;34mChecking dependencies...\033[0m\n"
-dep_ch "curl" "sed" "grep" "openssl" || true
+dep_ch "curl" "sed" "grep" || true
+case "$(uname -o 2>/dev/null)" in
+    *ndroid*) command -v openssl >/dev/null || die 'Program "openssl" not found. On Termux, install with: pkg install openssl-tool' ;;
+    *) dep_ch "openssl" || true ;;
+esac
 [ "$skip_intro" = 1 ] && (dep_ch "ani-skip" || true)
 dep_ch "fzf" || true
 case "$player_function" in
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.11.0"
+version_number="4.12.0"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -511,7 +511,7 @@ done
 [ "$use_external_menu" = "1" ] && multi_selection_flag="${ANI_CLI_MULTI_SELECTION:-"-multi-select"}"
 [ "$external_menu_normal_window" = "1" ] && external_menu_args="-normal-window"
 printf "\33[2K\r\033[1;34mChecking dependencies...\033[0m\n"
-dep_ch "curl" "sed" "grep" || true
+dep_ch "curl" "sed" "grep" "openssl" || true
 [ "$skip_intro" = 1 ] && (dep_ch "ani-skip" || true)
 dep_ch "fzf" || true
 case "$player_function" in
--- a/ani-cli
+++ b/ani-cli
@@ -208,13 +208,35 @@ select_quality() {
     episode=$(printf "%s" "$result" | cut -d'>' -f2)
 }
 
+decode_tobeparsed() {
+    blob="$1"
+    tmp="$(mktemp)"
+    ct="$(mktemp)"
+    key="$(printf '%s' 'SimtVuagFbGR2K7P' | openssl dgst -sha256 -binary | od -A n -t x1 | tr -d ' \n')"
+    printf '%s' "$blob" | openssl enc -d -base64 -A >"$tmp"
+    len="$(wc -c <"$tmp" | tr -d ' ')"
+    iv="$(dd if="$tmp" bs=1 count=12 2>/dev/null | od -A n -t x1 | tr -d ' \n')"
+    ct_len=$((len - 28))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by justchokingaround** [comment]
key is not likely to change soon, as it's hardcoded in the frontend bundle: `cdn.allanime.day/all/manga/a10191a.js`

**Review by 71zenith — APPROVED** [review]

**Comment by CoolnsX** [comment]
LGTM, however we need to mention additional dependency as well. I will refactor it in next patch as I am out of station.

**Comment by Derisis13** [comment]
> key is not likely to change soon, as it's hardcoded in the frontend bundle: `cdn.allanime.day/all/manga/a10191a.js`

Does the IV change though? If both are fixed (which'd be atupid, but we've seen worse), we can simplify the whole process to a single xor

**Review by on-ii — APPROVED** [review]

**Comment by justchokingaround** [comment]
> > key is not likely to change soon, as it's hardcoded in the frontend bundle: `cdn.allanime.day/all/manga/a10191a.js`
> 
> Does the IV change though? If both are fixed (which'd be atupid, but we've seen worse), we can simplify the whole process to a single xor

the IV is not fixed, it's the first 12 bytes of each encrypted blob, generated per-request

**Review by piganightis — APPROVED** [review]

**Review by port19x — APPROVED** [review]

**Comment by port19x** [comment]
Since this is verified as working by @Luuvydev and two randoms approving this PR, I'd prefer if this was merged today.
I'm not in the discord and don't plan to rejoin at this time, so I'll leave the timing up to your discretion and internal communication.

**Review by ramymoze — APPROVED** [review]

**Review by T-Bukovy — APPROVED** [review]

File: ani-cli [hunk 1]
```diff
@@ -208,13 +208,35 @@ select_quality() {
     episode=$(printf "%s" "$result" | cut -d'>' -f2)
 }
 
+decode_tobeparsed() {
+    blob="$1"
+    tmp="$(mktemp)"
+    ct="$(mktemp)"
+    key="$(printf '%s' 'SimtVuagFbGR2K7P' | openssl dgst -sha256 -binary | od -A n -t x1 | tr -d ' \n')"
```
Derisis13: this is a constant (a constexpr if you wish), as far as I can tell. Putting it to startup would make more sense logically than computing it here every time.

ani-cli
```diff
@@ -208,13 +208,35 @@ select_quality() {
...
+    printf '%s' "$blob" | openssl enc -d -base64 -A >"$tmp"
```
Derisis13: GNU coreutils has a program for base64 decoding (called base64, we've used in V2 or V3). I'd prefer its usage over openSSL as it decreases our dependency on openSSL.

ani-cli
```diff
@@ -208,13 +208,35 @@ select_quality() {
...
+    len="$(wc -c <"$tmp" | tr -d ' ')"
+    iv="$(dd if="$tmp" bs=1 count=12 2>/dev/null | od -A n -t x1 | tr -d ' \n')"
+    ct_len=$((len - 28))
+    dd if="$tmp" bs=1 skip=12 count="$ct_len" 2>/dev/null >"$ct"
```
Derisis13: I really don't like the whole operation of splitting a string (or binary blob) by writing to a file and `dd`-ing from it. Firstly, `dd` can read from STDIN (it does that by default), which I'd prefer for speed's sake (I think it has been demonstrated that people run this script on the crappiest hardware).
Second, the `ct_len` variable seems one billion per cent redundant, not setting count to anything copies the entire content of the input from `skip` to EOF.

My proposed solution to save some lines:
```sh
blob_raw=$(printf '%s' "$blob" | openssl enc -d -base64 -A)
ctr="$(printf '%s' "$blob_raw" | dd bs=1 count=12 2>/dev/null | od -A n -t x1 | tr -d ' \n')00000002"
ciphertext="$(printf %s "$blob_raw" | dd bs=1 skip=12 2>/dev/null)"
plaintext="$(printf '%s' "$ciphertext" | openssl enc -d -aes-256-ctr -K "$key" -iv "$ctr" -nosalt -nopad 2>/dev/null)"
```
seems much simpler for me and also significantly less steps (250% less lines!!! /j) and no file IO. I haven't tested it though, so it's probably wrong somewhere...

Derisis13 (changes requested):
Nice work on the reverse-engineering!

Please elaborate on my comments and see my suggestion on how to do the decryption more concisely and without temp files. Once these are over, I'll do some testing (I have some catching up to do). Any test cases that reliably need the decoding? (I'd also like list them in the PR template, because this is bound to stay)

Galenoo (approved):

ani-cli
justchokingaround: shell variables can't hold raw binary data since `$()` strips null bytes, which corrupts the iv extraction. i kept one temp file for the decoded blob so that dd can read the binary from disk

justchokingaround:
> Nice work on the reverse-engineering!
> 
> Please elaborate on my comments and see my suggestion on how to do the decryption more concisely and without temp files. Once these are over, I'll do some testing (I have some catching up to do). Any test cases that reliably need the decoding? (I'd also like list them in the PR template, because this is bound to stay)

i just use lain, but anw the decryption applies to all animes anw, so use whatever you want, it doesn't matter

ani-cli
Derisis13: Oh wow... Well that's something worthy of a short comment

Derisis13:
Just tried it on termux: openssl is weird. They separated it into two packages: openssl and openssl-tool. The CLI utility is in openssl-tool. This needs to be documented in the readme, and maybe as a case in dep_ch if we want to minimize the obvious question of "I installed openssl but script asks for openssl"

justchokingaround:
> Just tried it on termux: openssl is weird. They separated it into two packages: openssl and openssl-tool. The CLI utility is in openssl-tool. This needs to be documented in the readme, and maybe as a case in dep_ch if we want to minimize the obvious question of "I installed openssl but script asks for openssl"

done

Derisis13 (approved):
You're clear for merging

oilmine:
even after the fix, it still doesnt work for me

(nvm, do `sudo ani-cli -U`  (found from #1651) if it doesnt work for someone at first as the new version has only been published for Termux rn)

PR #1656, opened 2026-04-18
fix ani-cli

ErenYeager01001A:
## Problem
ani-cli was returning:
"Specified quality not found, defaulting to best
Episode is released, but no valid sources!"

This was caused by newer responses including a `tobeparsed` encrypted payload that was not being handled.

## Solution
This PR adds support for decoding the `tobeparsed` field by introducing a decryption step, allowing extraction of valid episode source URLs.

## Key Changes
- Added `decode_tobeparsed()` function to decrypt AES-256-CTR encoded responses
- Detects presence of `tobeparsed` in API response
- Extracts and decodes source URLs properly

## Result
- Restores working episode links
- Fixes "no valid sources" error across affected anime

## Verification
- Tested with multiple anime titles
- Confirmed streams work correctly via mpv

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="4.11.0"
+version_number="4.12.0"
 
 # UI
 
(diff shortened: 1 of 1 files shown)


oilmine:
i have been experiencing this very issue

nwroblewski:
https://github.com/pystardust/ani-cli/pull/1632 - there is change to POST merged and I feel like the PR description is not really accurate since you didn't change the HTTP method in the request rather than added decryption of the response. Please be more accurate now and in the future

ErenYeager01001A:
> #1632 - there is change to POST merged and I feel like the PR description is not really accurate since you didn't change the HTTP method in the request rather than added decryption of the response. Please be more accurate now and in the future

Thanks for pointing this out, I appreciate it.

You're right — the PR description was not accurate. The POST request change was already introduced earlier, and my change actually focuses on handling the `tobeparsed` encrypted response by adding a decryption step.

I’ll update the PR description to reflect this correctly. Thanks again for the clarification.

ErenYeager01001A:
> i have been experiencing this very issue

Thanks for mentioning this.

I've pushed a fix that handles the `tobeparsed` response, which should resolve the "no valid sources" issue. Feel free to try it and let me know if it works for you.

vazqurog:
> > i have been experiencing this very issue
> 
> Thanks for mentioning this.
> 
> I've pushed a fix that handles the `tobeparsed` response, which should resolve the "no valid sources" issue. Feel free to try it and let me know if it works for you.

Had the same issue. its fixed now after installing 4.12

Derisis13:
Literal duplicate of #1650.

PR #1663, opened 2026-04-21
docs: bump min version to 4.12 for packaging status

port19x:
# Pull Request Template

## Type of change

- [x] Documentation update

## Description

the min version in that repology widget is used to indicate that while a version above it may not be latest (green), it still works and has no security issues (light red).
A version below the min version instead turns to a saturated red.
After a long peace of stability, a full outage triggered 4.11 and 4.12.
By bumping this I hope to subtly nudge users and package maintainers to update faster and communicate that state.
This is not urgent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ A cli to browse and watch anime (alone AND with friends). This tool scrapes the 
 <h1 align="center">
 	Showcase
 </h1>
-I'm blue abedee abedai
+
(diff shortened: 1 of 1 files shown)


port19x:
I'm still half asleep, it's 4.12 that I bumped to.

port19x:
I added two nonsense commits, so squash and merge corrects the commit message.

Derisis13:
I suggest we wait with this until things settle. Or update it every PR, like version number.

Derisis13 (approved):
good to go

Issue #1693, opened 2026-04-27
Add `--profile` to use a selected profile when using mpv


germaniuss:
**Is your feature request related to a problem? Please describe.**

Currently there is no way to change mpv options specifically for anicli and I believe this is by design and agree with the philosophy since adding global options such as suggested `--fullscreen` may or may not work in all players. However the problem still persists that setting mpv options specifically for ani cli is currently impossible.

**Describe alternatives you've considered**

We currently have one of two options:
1. Set mpv.conf globally for all instances of mpv. This is not ideal since it is the core of the problem, and modifying mpv.conf every time I want to use it outside ani-cli is not great. In general no program should require to constantly modify configs for outside programs, even if it's a dependency.
2. Modify the ani-cli script. This is also not great since any update to ani-cli using `ani-cli -U` will automatically undo all your changes and you will have to redo them again. While creating an mpv profie and manually adding `--profile` in the script every time its updated does not take that much time it is still pretty annoying.

**Describe the solution you'd like**

I propose a middle ground solution, where a flag called `--profile` is added and all calls to the mpv player add the parameter `--profile` in the script. This way there is no ambiguity where this option can be used since it's mpv specific and not something like `--fullscreen` which may or may not work with different players.

Furthermore, there already are mpv specific features such as `--skip` so I do not believe adding another one goes against what is already established nor would be a maintenance burden.

germaniuss:
If you're interested I can make a PR since this will take at most 15 minutes to implement.

germaniuss:
Just as a note this issue was triggered by the update for the recent "no valid sources" issue. For those updates I am very grateful.

port19x:
Rejected as explained in the PR

PR #1694, opened 2026-04-27
feat: Add -p/--profile option

germaniuss:
Add -p/--profile option that directly sets the mpv --profile command line variable. Haven't really tested everything but it shouldn't affect most things anyway. Seems to work on my PC with `--no-detach` and a custom mpv profile. Not sure about the procedure here to bump version, I don´t want to merge this anyway until the other current issues are solved since I want to rebase on those.

# Pull Request Template

## Type of change

- [ ] Bug fix
- [X] Feature
- [ ] Documentation update

## Description

Solves the issue #1693 

## Checklist

- [x] any anime playing
- [ ] bumped version
---
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-e` (select episode) aka `-r` (range selection) works
- [ ] `-S` select index works
- [ ] `--skip` ani-skip works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -62,6 +62,8 @@ help_info() {
         Select nth entry
       -q, --quality
         Specify the video quality
+      -p, --profile
+        Use the specified mpv profile to use (mpv only)
       -v, --vlc
         Use VLC to play the video
       -V, --version
@@ -326,6 +328,7 @@ download() {
 play_episode() {
     [ "$log_episode" = 1 ] && [ "$player_function" != "debug" ] && [ "$player_function" != "download" ] && command -v logger >/dev/null && logger -t ani-cli "${allanime_title}${ep_no}"
     [ "$skip_intro" = 1 ] && skip_flag="$(ani-skip -q "$mal_id" -e "$ep_no")"
+    [ "$profile" != 0 ] && profile_flag="--profile=$profile"
     [ -z "$episode" ] && get_episode_url
     # shellcheck disable=SC2086
     case "$player_function" in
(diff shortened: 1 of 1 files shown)


justchokingaround:
no.

just change the value of the env var `ANI_CLI_PLAYER`

germaniuss:
@justchokingaround While I do understand not adding arbitrary mpv options to ani-cli, changing the env variable does not seem like the best solution. The reason I wanted to add this is precisely to not have to add arbitrary changes to the script and instead modify a profile in mpv.conf. In general what are the guidelines to adding new options since this seems very unintrusive to me at least.

justchokingaround:
u declare the player env var outside of the script... I don't get what u mean by arbitrary changes to the script

germaniuss:
Sorry I did not make it very clear. I meant that when adding the --profile option it did not seem as arbitrary to me as say adding --fullscreen or say --speed. I understand that env variable are set outside but that seems very unergonomic and adding a --profile option actually makes it easier to set options for ani-cli in mpv.conf and clearer for the user that the indended way of using ani-cli is by setting modifying profiles in mpv.conf.

germaniuss:
With regards to what the guidelines are for adding new options, where is the line drawn? Is any mpv option banned? But if that is the case why is say --skip which is mpv specific used? How does one determine what can or cannot be added when it comes to features for specific backends?

justchokingaround:
@port19x @Derisis13 I'll let u guys decide. I understand the idea but I'm not really sure what my stance is

port19x:
If ani-skip was proposed today I would probably reject it.
Too difficult/tedious to set up and a moderately trustworthy outside dependency we have no oversight over.

The distinction between environment variables and flags is supposed to subtly communicate something:
By using environment variables, which are documented in the man page, you're a proficient linux user.

It's a grey area where we make a concession to power users, so they don't get the full integration as mpv, vlc, iina and syncplay, but also don't have to edit the script themselves.
If this wasn't a feature, we'd get even more than our monthly new player PR.

PR #1702, opened 2026-05-03
feat: Add an method to check the version, and prompt user to update ani-cli.

Jay-1409:
## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description
helps users against no ep released issues cased due to older version of ani-cli

**Version 4.13.0**
image

**Version 4.14.0**
image

## Checklist

- [x] any anime playing
- [ ] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -122,20 +122,22 @@ update_script() {
 # compare two versions: returns 1 if v1>v2, 0 if equal, -1 if v1<v2
 #compare at each level
 ver_compare() {
-    v1="$1"
-    v2="$2"
-    awk -v a="$v1" -v b="$v2" 'BEGIN {
-        n1 = split(a, A, ".");
-        n2 = split(b, B, ".");
-        n = (n1>n2)?n1:n2;
-        for(i=1;i<=n;i++){
-            ai = (i in A) ? A[i] : 0;
-            bi = (i in B) ? B[i] : 0;
-            if(ai+0 < bi+0){ print -1; exit }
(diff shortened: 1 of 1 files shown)


Derisis13:
Even though lately it has been the case, we don't always release a new version due to breakage. 
With that said, the request is async, so it doesn't slow the script down, and it's not blocking execution, only showing a warning. I like these.
What I don't like is awk (not cross compatible), and that it adds too many lines compared to the function it provides. Unless it can be done in a more conscise way,  I can't fully support it.

Jay-1409:
@Derisis13

Thanks for the feedback,
it really made me think a lot, and I implemented the recent commit. 
Could you have a look at it and provide any feedback?  

: )

port19x:
Idk, I don't like it

port19x:
The code is way too long and you're evidently still using awk.

But beyond the code, I find such warnings rude, especially considering we're in several package repositories where that should be the primary way to update.
`-U` is a fallback because ani-cli can experience real breakage and package maintainers tend to be on the slow side, plus for manual installs.

Jay-1409:
Right,
Thanks for the feedback, 

appreciate it!

PR #1755, opened 2026-06-26
feat: add mpv playlist support

### LINKED ISSUE #1568: better multi-ep support with syncplay
Syncplay integrates ani-cli's auto-next / next episode buttons poorly.
To play the next episode without breaking, one has to manually close the player and syncplay, and only after that start the next episode, rejoining the room / server etc.

Next button / going next episode adds next link to syncplay playlist, then plays next playlist entry, without leaving the room. Same thing should be applied to similar features like replay etc.

**Describe alternatives you've considered**
Write entire playlist to a temp file, tell syncplay to read it (--load-playlist-from-file) and completely remove next episode logic (with the -s flag) and use the syncplay one. Potential issue here could be that the change quality feature would no longer exist.

mbpowers:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Documentation update

## Description

Added an option `-p` or `--playlist` which adds episodes to a single mpv instance's playlist instead of spawning a new instance each episode. A new episode is only added to the mpv playlist when you have reached the last playlist entry and you haven't reached the end of the range i.e. `get_episode_url` will only run as needed to avoid spamming.

I neglected to update the README/Manpage yet because I am wondering about the implications of me adding the dependency `socat` for listening to the mpv IPC unix socket. Is this dependency justified? I am not familiar with how ani-cli works Windows and it seems like won't support this unix socket approach. In that case using `netcat` or something for tcp sockets could be best? Or would this be fine as unix only for now? And if no dependencies are justified for this then it could be possibly reworked with fifos and an mpv lua script, but since they block I think that would be more complicated with the current code structure, and that would still be unix only.

Anyways, I wanted to ask before messing with it more as I am happy with where this is at for my own purposes. Let me know if there are any suggestions or required changes

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- [x] `--skip` ani-skip works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,8 +51,15 @@ A cli to browse and watch anime (alone AND with friends). This tool scrapes the 
 
 ## Fixing errors
 
-If you encounter `No results found` (and are sure the prompt was correct) or any breaking issue, then make sure you are on **latest version** by typing
-`sudo ani-cli -U` to update on Linux, Mac and Android. On Windows, run `ani-cli -U`.
+If you encounter `Blocked by cloudflare. Try installing curl-impersonate` then install `curl-impersonate` from your respective package manager.
+If it is not available, then download from their [github](https://github.com/lwthiker/curl-impersonate) by running the following commands.
+
+```sh
+curl -LO "https://github.com/lwthiker/curl-impersonate/releases/download/v0.6.1/curl-impersonate-v0.6.1.x86_64-linux-gnu.tar.gz"
+sudo tar xf curl-impersonate-v0.6.1.x86_64-linux-gnu.tar.gz -C /usr/local/bin
+```
+
+For any other breaking issue, then make sure you are on **latest version** by typing `sudo ani-cli -U` to update on Linux, Mac and Android. On Windows, run `ani-cli -U`.
 If after this the issue persists then open an issue.
 
 ## Install


port19x:
This has been a desired feature for multi-ep handling in the past, but I can tell you right now that we're not merging a +89-15 diff

port19x:
If you're willing, playlist support for syncplay would also be valuable

mbpowers:
So the size alone would be disqualifying? What size would be suitable?

A decent bit of the diff comes from extracting the title into "$media_title" and extracting the play_episode mpv case into `play_mpv()`. The actual added code is closer to ~60 lines and is mostly contained inside `mpv_playlist_init()` and `mpv_playlist_wait()` which only ever get called when `-p` is set.

The changes outside of extracting `$media_title` into a variable, `mpv_playlist_init()` and `mpv_playlist_wait()` include:

in `cleanup()`
---
```bash
[ -n "$playlist" ] && rm -f "$MPV_SOCKET"
```
only affects `-p`

in `get_episode_url()`
---
```bash
pids=""
for provider in $providers; do
    generate_link "$provider" >"$cache_dir"/"$provider" &
    pids="$pids $!"
done
if [ -n "$playlist" ]; then
    for pid in $pids; do
        wait "$pid"
    done
else
    wait
fi
 ```
previously
```bash
for provider in $providers; do
    generate_link "$provider" >"$cache_dir"/"$provider" &
done
wait
```
which now filters wait to wait for just the pids of `generate_link` so that the function does not wait on `mpv` to exit to generate links. Only affects `-p`.

in the `play_episode()` mpv case becomes
---
```bash
mpv*)
	if [ -n "$playlist" ]; then
	    if [ -z "$mpv_playlist_pid" ] || ! kill -0 "$mpv_playlist_pid"; then
	        mpv_playlist_init
	    else
	        [ -z "$episode" ] && return
	        send_mpv_ipc "{\"command\":[\"loadfile\",\"$episode\",\"append-play\",-1,{\"force-media-title\":\"$media_title\"}]}" >/dev/null 2>&1
	        mpv_playlist_wait
	    fi
	else
	    play_mpv
	fi
	;;
```
where the previous code was extracted into `play_mpv`. Only affects `-p`.

at the end of `play_episode()`
---
```bash
[ -n "$mpv_playlist_quit" ] && ep_no=$(( ep_no - 1 )) && return
```
because with `-p` we always fetch one episode ahead, if you quit early it must be adjusted back. `$mpv_playlist_quit` is only set with `-p`.

in the episode range loop in `play()` to quit early
---
```bash
if [ -n "$mpv_playlist_quit" ]; then [ -n "$exit_after_play" ] && exit || return; fi
```
again, `$mpv_playlist_quit` is only set with `-p`.

---

So I don't think any of my changes affect anything outside this feature, although a bit of code was moved around.

Also the shellcheck errors were in the previous commit I believe and rely on variable splitting to work.

Let me know if there is a path forward. If so I will take a look at syncplay

port19x:
We have a strong preference for brevity and consistently apply this priority to choose what feature requests to accept or reject.
Code size is a more important factor than feature utility in this way.

This is because ani-cli is a finished product and well into the maintenance phase from a lifecycle standpoint, so if we never add another feature again it's still a success.
Shorter features or very valuable features do get considered, but you'd have to propose something extraordinary to warrant the 10% increase in code size that you propose.

It should be noted that modifications to the readme and the man page don't count towards this limit

mbpowers:
To explain why I think this feature is very desirable:
- It allows you to maintain mpv settings that are changed while watching (volume, playback speed etc.)
- Maintains a single mpv window so you do not need to mess with window positioning/swallowing behavior
- Next episode url is prefetched and queued so no delay
- Allows skipping forward/back episodes from within mpv and all episode urls stay cached in mpv

But either way, I understand the concern. Just let me know if there is something I can do to make it meet the requirements, like how much more concise it would need to be etc. cause I do think people would appreciate this feature, otherwise I am happy just using my personal fork.

Derisis13:
I don't want to completely say no to this feature, I beleive it could be rather useful; as a refactor. Making it another bolt-on feature would be coutraproductive in my opinion too.

The way I would've approached it would have been to spawn a player process at startup, and only terminate it on shutdown. Next, previous, replay and list would be handled by socket commands to this player process.
The obvious problem to this approach is that mpv (and variants) and vlc are completely different, it'd require lots of code to test for both. 

The issue was downloading. Implementing socket control for download would be ridiculous, but since downloanding was modified to quit after the requested episodes were downloaded, it can separate early in the control flow, and maybe not require too much extra code.

The other issue is android, @CoolnsX probably has the best idea if it'd work there or not. iSH is broken anyway, so I wouldn't sweat that.  

So my personal approach would be to get rid of the old way of playing in favor of the new, make it so that it works with VLC, see if it works out or not, and then merge after a ton of testing.

port19x:
Here is how I would envision playlist integration:
- No extra flag
- No ipcs, pids, sockets
- Integration with `-e` and `-r`
- mktemp (beware cross platform catches)
- mpv --playlist=/tmp/playlist.txt
- Do not document "mpv only". 90% of non-mpv players used are mpv based and will support mpv options as fundamental as --playlist.
- ca +30 loc acceptable without question, +50 if you need extensive catches for cross plattform and syncplay support

mbpowers:
> No ipcs, pids, sockets

We need some form of communication otherwise we would need to load all the urls upfront. Is there some form of communication you would suggest? No IPCS, pids or sockets due to windows compatibility? 

It seems like mpv only supports unix sockets (on unix) and named pipes (on windows). I am already using `socat` for unix and I believe `WinSocat` could work on windows, but it has a slightly different syntax.

The only other option I see is mpv allows for lua scripts that have hooks that can run on events and we could just read from a file or something, but then it is no longer bidirectional and would require polling. If we are only using it for adding episodes though this wouldn't be that bad and we wouldn't need to poll very frequently.

Or were you thinking we load the urls upfront or?

> Integration with -e and -r

This is currently supported, is how it is working now fine? With just -e it works as normal, whenever a range is set it will keep adding episodes to the playlist until the range is concluded. Both -e and -r return to the menu once they conclude and you can select next episode or a new range or whatever as normal.

port19x:
> otherwise we would need to load all the urls upfront

I do find this to be desirable, that would also easily check the syncplay compatibility box, a feature I would 100% use on the regular.

If -e and -r work as you described than you're already done with that part, sounds promising.

port19x:
Of course one limitation we have to test is how long the links remain valid.
If they expire in less than an hour, fetching them all in advance won't work.
If they last for 6 hours that would be ideal

mbpowers:
> I do find this to be desirable

Ah okay that definitely simplifies things. I was worried about spamming the provider.

I think they last quite a while because I often leave an anime paused on another window for, I want to say even days at points, but definitely many hours and I have never noticed it fail to load from where I left off. I will do some tests.

edit: thinking about it though the episode was just fully loaded probably

The only other concern I have with loading up front is you need to know how many episodes you want to watch beforehand if you want it contained to a single playlist. I often do `-r 0..-1` so that I don't need to think about the end episode and I can just close out when I want. Obviously it would not be great to preload them all in that case. Especially with 1P lol. As a compromise we could have a configurable maximum fetch amount so you can load say 10 at once, they go into a playlist, once done watching those mpv closes, we load 10 more, it reopens, and so on.

Maybe not a concern if the links last long enough though.

port19x:
Yes. A sensible maximum fetch amount would be 30. That's a little more than a standard 24 episode anime season.
This should be configurable with an environment variable.
Let's call it `ANI_CLI_PLAYLIST_MAX`.

You can do some testing if that default of 30 would be rate limited.
If so, we can reduce it to 12.
Also keep in mind that these requests should be done concurrently if possible, which is easily achieved with the `&` operator.

mbpowers:
Here are my findings on playlist functionality across players. Seems like m3u8 is widely supported. If anyone can test the basic functionality of loading multiple media on iina and iSh that would be helpful as I don't have a device to test.

- android_mpv (tested, works)
	- pass m3u8 using am
	
- android_vlc (tested, works)
	- pass m3u8 using am
	
- mpv linux (tested, works)
	- pass urls as args or m3u8
	
- vlc linux (tested, works)
	- pass urls as args or m3u8
	
- Syncplay (untested, should work)
	- one file per line only, no m3u/m3u8
	- https://github.com/Syncplay/syncplay/issues/408
	
- iina (untested, cannot test, should work)
	- requires m3u8 or piping to stdin
	- pass as args doesn't work
	
- iSH (untested, cannot test, should work)
	- may require m3u8?
	- does the click to open work with m3u8?
	- can we use `open "vlc-x-callback://x-callback-url/stream?url=ENCODED_VIDEO_URL"`?
	
- catt (untested, cannot test, probably won't work)
	- m3u8 not working
	- https://github.com/skorokithakis/catt/issues/490
	- https://github.com/skorokithakis/catt/issues/360

Mpv and vlc both use the episode names from the m3u8 tags like `#EXTINF:1500,One Piece Episode 1`. It seems syncplay intentionally does not support this so episode titles will be lost there in playlist mode.

Also just trying the new provider today, the links are still valid after 4 hours.

Am I missing any players?

port19x:
Very interesting data.
You should probably redo the PR from scratch based on the current v5 master

port19x:
If you comment on the linked issue I can assign you

mbpowers:
Okay working on this a bit and thinking about it more, without the inter-process communication we wont know if/when the user quits the playlist, which means they won't be able to quit, change quality, and resume etc. and their history will be messed up if they just want to stop watching early.

Mpv allows for lua scripting and we can attach a function to a hook to write the current episode in the playlsit to a tmpfile. This would remove the need for platform specific communication such as sockets etc. This still is much simpler than the previous approach as the communication is only one way, we still fetch a number of episodes up front, the player just writes the last episode watched and we read that file when needed.

Vlc seems to also have lua extensions but I haven't looked into it fully yet. Since iina is on top of mpv the same lua functionality should be supported as well. So this would mean two small lua scripts, one for mpv, one for vlc.

If a player doesn't support reporting the playlist position (which should only be catt?) we can always fall back to assuming it was consumed fully, but I wouldn't be happy with this being the standard.

Thoughts? I will look into this more if there are no hard objections

mbpowers:
for example:
```bash
    mpv_script="mp.register_event('shutdown', function ()
  local pos = mp.get_property_number('playlist-pos')
  os.execute('echo '..pos..' > $tmpdir/episode')
end)"
    echo "$mpv_script" > "$tmpdir/script.lua"

    mpv --script="$tmpdir/script.lua" "$m3u8"
```

mbpowers:
vlc would require polling and look like
```bash
echo "while true do
	local item = vlc.input.item()
	if item then
	    local uri = item:uri()
	    if uri ~= last then
	        os.execute('echo \"'..uri..'\" > $tmpdir/episode_url')
	    end
	    last = uri
	end
	vlc.misc.mwait(vlc.misc.mdate() + 10000000)
end" > "$vlc_script_dir"

vlc -I luaintf --lua-intf script "$m3u8"
```
where `$vlc_script_dir` would need to be the corresponding directory that vlc expects. I am not exactly sure what dir that would be for iSH or termux though or if we would have permissions to place the script there...

mbpowers:
It appears that vlc does not support lua plugins/extensions on android or ios

port19x:
You definitely raise a valid point regarding how playlists interplay with history.
I don't think that IPC is the answer, perhaps process polling is a thing?
Can we check what file is being played in MPV through a `pgrep mpv` or similar and then count changes to that?
Definitely some open questions here

mbpowers:
The pid doesn't change on the instace since it is the same player. I also don't see any way of getting the media title or current media playing that isn't tied to what window is open and then would be different between X11, wayland even ignoring other OSes/platforms. Couldn't find anything in /proc/ but if anyone knows better lmk. Mpris is a thought but that's another dependency and linux only.

Also I misspoke before, this wouldn't be IPC, its just giving mpv/vlc a small piece of lua for them to execute to save their playlist state to a tmpfile which ani-cli can read whenever it needs (mpv provides a hook on shutdown and vlc does polling).

The only options from everything I've seen are either give them lua to execute or communicate over IPC with unix sockets/named pipes.

Overall it seems that communication with vlc on ios/android will not be possible with either lua or IPC, mpv on android would require lua, and the rest would work on either afaik.

Imo we go with lua and decide whether the unsupported platforms should assume all loaded episode were watched or just disallow playlist mode.

mbpowers:
@port19x should I proceed in the lua direction then or is that solution not satisfactory?

port19x:
@mbpowers stop working on the feature, it would bloat ani-cli greatly and introduce problems with history, especially as I don't want IPC

PR #1772, opened 2026-07-08
fix: handle AllAnime aaReq token for episode sources

bl33dz:
AllAnime changed its GraphQL API. Episode source requests now require an
AES-256-GCM `aaReq` token in the `extensions` object and must be sent as GET
with query parameters. POST requests fail with `AA_CRYPTO_MISSING`.

#### Changes:
- `process_response()` — Decrypt `tobeparsed` with AES-256-GCM instead of
  AES-256-CTR (openssl `enc` doesn't support AEAD, so uses Node.js `crypto`)
- `get_episode_url()` — Generate dynamic `aaReq` token via the same XOR-derived
  key and include it in the GraphQL extensions; removed dead POST fallback
- `allanime_key` — XOR-derived key from two static secrets (matching the
  official site's build) instead of SHA256(`Xot36i3lK3:v1`)

The aaReq token is time-bound (5-minute window) and encrypted with AES-256-GCM
using the same key, matching the exact protocol the AllAnime website uses.

No new runtime dependencies — Node.js `crypto` is built-in. `nodejs` added to
dependency list in README and dep checker.

#### Before Fix:
image

#### After Fix:
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,11 @@ For players you can use the apk (playstore/fdroid) versions of mpv and vlc. Note
 pkg install openssl-tool
 ```
 
+**Note:** `nodejs` is also required for AllAnime source decryption. Install it with:
+```sh
+pkg install nodejs
+```
+
 **Important Note:** To get all providers working with android MPV, Please follow below steps:
 - Run this command and allow storage permissions:
 ```sh
@@ -527,7 +532,8 @@ apk del grep sed curl fzf git aria2 ffmpeg ncurses
 - yt-dlp - m3u8 Downloader
 - ffmpeg - m3u8 Downloader (fallback)
 - fzf - User interface
-- openssl (for decrypting encrypted video sources; on Termux, the CLI is in the `openssl-tool` package)
+- openssl (for decrypting Filemoon video sources; on Termux, the CLI is in the `openssl-tool` package)
+- nodejs (for decrypting AllAnime video sources)
 - ani-skip (optional, for auto-skipping anime intros)
 - patch - Self updating
 
--- a/ani-cli
+++ b/ani-cli
@@ -239,14 +239,21 @@ process_response() {
         return 0
     fi
 
-    tmp="$(mktemp)"
-    printf '%s' "$1" | sed -nE 's|.*"tobeparsed":"([^"]*)".*|\1|p' | base64 -d >"$tmp"
-    file_size="$(wc -c <"$tmp")"
-    iv="$(dd if="$tmp" bs=1 skip=1 count=12 2>/dev/null | od -A n -t x1 | tr -d ' \n')"
-    ctr="${iv}00000002"
-    ct_len=$((file_size - 13 - 16))
-    dd if="$tmp" bs=1 skip=13 count="$ct_len" 2>/dev/null | openssl enc -d -aes-256-ctr -K "$allanime_key" -iv "$ctr" -nosalt -nopad 2>/dev/null
-    rm -f "$tmp"
+    printf '%s' "$1" | sed -nE 's|.*"tobeparsed":"([^"]*)".*|\1|p' | node -e "
+const crypto = require('crypto');
+const key = Buffer.from('$allanime_key', 'hex');
+let d = '';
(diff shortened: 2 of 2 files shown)


candrapersada:
but have to install node and doesn't work in deno
but if use `deno eval` work without `node -e`
and Yt server missing from ani-cli
and some anime don't have `mp4upload` as server
image
or 
image

Rikaisan:
I personally don't like the inclusion of a JS runtime as a dependency, as it can lead to ani-cli packages being prone to all of the JS-targetting malware worms, are there no other alternatives?

guidoaguiar:
i had to force this PR on my machice to make it work, it was not finding any media to play

skaarl-user:
> I personally don't like the inclusion of a JS runtime as a dependency, as it can lead to ani-cli packages being prone to all of the JS-targetting malware worms, are there no other alternatives?

aren't those only on NPM? (in the same sense of the AUR incident)
using just native node.js libs can't bring worms by themselves without a zero day exploit, on Node.js itself, different than supply chain attack when you use npm/bun install deps.

correct me if I'm wrong tho, I always use --ignore-scripts and minimize deps footprint and avoid adding big/new ones.

candrapersada:
> > I personally don't like the inclusion of a JS runtime as a dependency, as it can lead to ani-cli packages being prone to all of the JS-targetting malware worms, are there no other alternatives?
> 
> aren't those only on NPM? (in the same sense of the AUR incident) using just native node.js libs can't bring worms by themselves without a zero day exploit, on Node.js itself, different than supply chain attack when you use npm/bun install deps.
> 
> correct me if I'm wrong tho, I always use --ignore-scripts and minimize deps footprint and avoid adding big/new ones.

or use [deno](https://deno.com/)

skaarl-user:
> or use [deno](https://deno.com/)

I've used it before on my personal projects, but the matter of my question is about the package manager, how safer is it than other options? what does it do different in regards to the supply chain attacks I mentioned on NPM

I know they bundle a bunch of useful tools directly, but looking into the NPM compromises am I missing something?

tnx for recommendation anyways, I'm already about 6 months away from js world cause I took the decision of diving into Linux and I am now learning the low level stuff I ignored during years

bcain2121:
Thanks for this fix it worked for me.

Hey I am a newby here 100% a vibe coder, I was wondering if someone with more experience could explain how they reverse engineered the allanime website to create this app its amazing. I was wondering if its possible to do this basically with any streaming website.

bl33dz:
> but have to install node and doesn't work in deno but if use `deno eval` work without `node -e` and Yt server missing from ani-cli and some anime don't have `mp4upload` as server image or image

I choose to use node because I believe its more familiar to most user.

bl33dz:
> I personally don't like the inclusion of a JS runtime as a dependency, as it can lead to ani-cli packages being prone to all of the JS-targetting malware worms, are there no other alternatives?

Theres a reason why i choose `node`, because crypto is built-in module and requires no npm dependencies. I already tried using `python` but it would require dependencies like `cryptography` or `pycryptodome`.

I also considered the `openssl` but `openssl enc` doesn't support AES-GCM for encryption/decryption as stated in OpenSSL Docs: https://docs.openssl.org/3.3/man1/openssl-enc/#supported-ciphers
> This command does not support authenticated encryption modes like CCM and GCM, and will not support such modes in the future.

mbpowers:
I noticed the `allanime_key`, `buildId` and `epoch` are hardcoded. In #1774 I was able to scrape these from their scripts. 

The `mask` and `partB` (XORed to generate the key) changed yesterday though key generated from them remained the same... which is kinda strange

The `buildId` has already changed to `12` as well but it seems to have no affect what it is set to.

Also the `epoch` of `4128` hasn't changed I'm not entirely sure what that indicates or if it will change. I would've expected that counts days but. I did try `4129` and `4127` and that causes it to fail though.

So it seems like hardcoding is relatively stable for now but these are my findings.

port19x:
No nodejs, but thanks for providing a workaround during our outage

PR #1774, opened 2026-07-08
fix: new aareq flow

mbpowers:
# Pull Request Template

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation update

## Description

A version of the fix that uses python and parses the all_anime key from scripts instead of hardcoding. Not sure if this is best way but in case it is useful to anyone.

The key did stay the same since yesterday despite the fact the mask and build_id changed (the two components XORed to make the key), so maybe this is unnecessary if the key stays consistent.

The script containing the mask and buildId changed structure since yesterday in a way that broke the regex. Hopefully this continues to work but it may be fragile.

If we do want to scrape this and it only changes daily, we may want to cache this instead of checking each time we get_episode().

## Checklist

- [ ] any anime playing
- [ ] bumped version
---
- [ ] `-d` downloads work
- [ ] `-s` syncplay works
- [ ] `-q` quality works
- [ ] `-v` vlc works
- [ ] `-e` (select episode) aka `-r` (range selection) works
- [ ] `-S` select index works
- [ ] `--skip` ani-skip works
---
- [ ] Readme is up to date
- [ ] Man page is up to date

## Additional Testcases

- The safe bet: One Piece


DarioTD:
Hi can you add in deps python-pycryptodome? It wasn't installed. After I did. It worked 👍

port19x:
No Python, but thanks for providing a stopgap solution during our outage

PR #1777, opened 2026-07-09
feat: tty render modes

whyshouldibother:
# feat: tty render modes

## Type of change

- [X] Feature

## Description

This version of ani-cli adds tty support for those users who don't want to use the terminal through a GUI. It adds two modes --tty 1 or --tty 2 which works with mpv video outputs as tct and drm respectively. 
So why would you even want to use ani-cli from pure tty?
- Well sometimes when i update something or try to do something i am not well informed about and then delete my DE/WM i would still like to be able to enjoy anime although that might not be the best way to enjoy it. 
- Or maybe someone just likes to watch anime as ascii art from their tty instance. 
-  Or, check out anime from ssh sessions where you would like to download them but can't verify them as you don't have a vnc setup and you just need to see if its the right one.
### Disclaimer
- I have tested this on kitty as well as my tty. I am currently on arch linux, with kernel 7.0.12-arch1-1. I assume the functionality is going to be same on other setups as well as it is depended on mpv rather than the terminal and operating system's behaviour to my knowledge.
- As of creating this PR the links are not working right now which is not caused my changes as they are not working properly even on a clean instance from github. I had checked it on my previous PR where it was working as intended and i haven't made any changes that should affect it and only have changed minor things which is only done to be compliant with the different CI/CD checks. 

This is the second PR i am requesting as my previous ones were failing the CI/CD checks which i have tried to fix in this one. 

## Checklist

- [X] any anime playing
---
- [X] `-d` downloads work
---
- [X] Readme is up to date
- [X] Man page is up to date

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -59,6 +59,10 @@ help_info() {
         Delete history
       -l, --logview
         Show logs
+    -t, --tty
+        Use tct or drm backend to use it in tty
+        tty 1 = tct (default)
+        tty 2 = drm
       -s, --syncplay
         Use Syncplay to watch with friends
       -S, --select-nth
--- a/ani-cli
+++ b/ani-cli
@@ -427,10 +427,7 @@ play_episode() {
                         tty_flags="--vo=tct --really-quiet"
                         ;;
                     2)
-                        tty_flags="--vo=gpu --gpu-context=drm"
-                        ;;
-                    *)
-                        tty_flags="--vo=tct --realy-quiet"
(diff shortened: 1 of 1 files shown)


port19x (changes requested):
Too large to an unsalvagable degree imo.
This needs to be possible via our provided idioms of --no-detach and ANI_CLI_PLAYER.
The fact that you have a large chunk in our mpv*) case where you add tty_flags proves that it would work using ANI_CLI_PLAYER

PR #1779, opened 2026-07-09
fix: Add AES-256-GCM encryption

### LINKED ISSUE #1763: Episode is released, but no valid sources!
ani-cli -V
4.14.1
  ani-cli
Checking dependencies...
Specified quality not found, defaulting to best
Episode is released, but no valid sources!

mine is still here

Image

71zenith:
# Pull Request Template

## Type of change

- [x] Bug fix

## Description

*ramble here*

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [x] `-d` downloads work
- [x] `-s` syncplay works
- [x] `-q` quality works
- [x] `-v` vlc works
- [x] `-S` select index works
- ❌ `--skip` ani-skip works
- ❌ `--skip-title` ani-skip title argument works
- ❌ all providers return links (not necessarily on a single anime, use debug mode to confirm)
---
- [x] Readme is up to date
- [x] Man page is up to date

## Additional Testcases

- The safe bet: One Piece


candrapersada:
is the botan binaries without using vcpkg or WSL on windows?

michaelssingh:
Not sure what is going on over there but they reverted back to AES-256-CTR

port19x:
botan exists on mac, but this PR doesn't work yet
Screenshot 2026-07-12 at 15 53 09

71zenith:
Because they have reverted back to AES-256-CTR so the master branch has started working again.

Derisis13:
> Because they have reverted back to AES-256-CTR so the master branch has started working again.

Not entirely true, master works *most of the times*

candrapersada:
how to get `botan` binaries without using vcpkg or WSL on windows?

port19x:
I'm currently browsing the openssl github and documentation to see if I can make it work with openssl

port19x:
I give up.
But bad AEAD support is known to openssl and has several open issues, so there is hope for proper support in a couple years.

Since we surely have homies reading along with python or go scrapers, those do support this via python cryptography and gos standard library, at least alledgedly so.

Derisis13:
I think I can follow the AES-GCM standard to implement the GHASH, in another language, with basic arithmetics and logic. Then porting it to posix should be easy enough, right?

port19x:
@71zenith @justchokingaround Could it be that filemoon is just dead code rn?

71zenith:
> @71zenith @justchokingaround Could it be that filemoon is just dead code rn?

It is dead code yes

port19x:
Also I think there is an opportunity for a principled solution regarding multi-path dependencies, such as iina, mpv and now botan.

We could probably overload `dep_ch` in a backwards compatible way to catch some extra cases in there, such as alternative program names or specific paths.

rough idea:
image

Advantages:
- set programs once in the script
- easy failover dependencies, usable both for different programs (vlc, mpv) or different names (botan3, botan)
- platform handling

Disadvantage:
- not sure how to cleanly the easy ones, like curl. Calling dep_nmlz "curl" "curl" "curl" ... is a bit obnoxious, there needs to be some implicit handling, but I can design that.
- painful to do the first port

port19x:
```sh
dep_ch_failover() {
    [ -z "$1" ] && return 1
    prog=$(printf "%s" "$1" | cut -d "," -f 1)
    rest=$(printf "%s" "$1" | cut -d "," -f 2-)
    command -v "$prog" >/dev/null 2>&1 && printf "%s" "$prog" && return 0
    [ -e "$prog" ] && printf "%s" "$prog" && return 0
    dep_ch_failover "$rest"
}
```

example call
```sh
#this would come from function args
$1="mpv,vlc,/usr/bin/ffplay"
...
case "$(uname -a | cut -d " " -f 1,3-)" in
    *) player_function=$(dep_ch_failover "$1") || die "missing dependency: $1"
...
```

port19x:
See #1805 for my full implementation of `dep_ch_failover`, providing big improvements to our dependency handling.

Thx @justchokingaround for the quick merge!

port19x:
Btw if there was any doubt that this is impossible with openssl: 

> This command does not support authenticated encryption modes like CCM and GCM, and will not support such modes in the future.

[openssl-enc 3.3 docs - supported ciphers](https://docs.openssl.org/3.3/man1/openssl-enc/#supported-ciphers)
2015 openssl/openssl#471
2020 openssl/openssl#12220
2024 openssl/openssl#24421
2025 openssl/openssl#26670

We gotta use botan or become cryptographers ourselves 🙃

Minor hope: openssl/project#248

port19x:
[Botan availability:](https://repology.org/project/botan/badges)
- ✅ Linux (Arch, Debian, Fedora, etc.)
- ✅ MacOS Homebrew
- ✅ Android Termux
- ✅ iOS iSH (Alpine Linux apk)
- 🟡 Windows

Windows is available via `vcpkg`, but it would be better if we just compile it from source for windows in our CI pipeline.
As per randombit/botan#1633, they will not distribute a binary themselves. I can probably make a working CI build pipeline this evening, or on the weekend at the latest.

port19x:
Nvm that last commit message, `cut -s` is indeed posix, as per the [opengroup spec](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cut.html)

port19x:
Someone tell #support in the Discord I fixed the hanging after "checking dependencies".
No way noone ran into this, considering how several people probably still don't know about the botan dependency.
@71zenith @candrapersada

port19x:
Thx zen, #1806 claims we need to add
`allanime_query_hash="f4662f4b7510b26795dd53ef824a0bf1740fbbc5d1273fab18222ac831bca8d0"`
I'm not changing this until I'm home in 2-3 hours, but maybe you have the time

71zenith:
It is indeed the new query hash. Allanime does have a buffer window so the d05.... hash is still working for me. I was just about to scrape the query hash and write a basic workflow such as in anipy-cli. 
Would you prefer a gist or in repo. I would personally not like to dirty the repo.

port19x:
> Would you prefer a gist or in repo. I would personally not like to dirty the repo.

For what exactly?
If you mean for the botan windows binary, I'm considering moving the .exe into .assets.
our action would be .github/workflows/botan.yml and trigger on master pushes that change said botan.yml

71zenith:
For fetching the key, epoch, query_hash

port19x:
Do you think they will keep rotating?
I'm not sure how to deal with that actually.
Imo it makes sense to use github actions for this and write it to some file, possibly on a separate branch.
But then that breaks up the logic into ani-cli and a helper script, but possibly save on network IO, as we can request one github cdn file instead of possibly multiple requests to allanime.

Dhairya3391:
I think zen wants to point thise hardcoded values to cone from some other file or somewhere, so the script just reads values from it and repo doesn't get dirty, the values in the gist or wherever we get them from can be changed automatically ( some way like scraping it ) or manually pushing those changes there. Mainly for users to not have to update script repeatedly.

71zenith:
> Do you think they will keep rotating? I'm not sure how to deal with that actually. Imo it makes sense to use github actions for this and write it to some file, possibly on a separate branch. But then that breaks up the logic into ani-cli and a helper script, but possibly save on network IO, as we can request one github cdn file instead of possibly multiple requests to allanime.

I think the frequency of updates might be high because of the transition to mkissa, but I can see then changing the keys still often enough that hardcoding isnt viable.
Alternatively, we can fetch them at runtime and cache it locally, only refetching if we encounter an error relating to `AA_CRYPTO_` or at regular intervals(say 2-3 days) . This would require a bit of care to implement to not result in a infinite loop or refreshing more than needed.

port19x:
I mean we can profile this.
By default I'd prefer not caching the values on the file system and would rather regenerate them each time we start the script.

If we can derive key, epoch and query hash all from the same request, I'd definitely do it in script.
If we need a separate request for each, that would be 3 requests, in that case we should think about caching or prefetching them in a repo or sth.

port19x:
ilarp is my alt for now. I don't plan on staying in the discord indefinitely, but I gotta be in there during these turbulent times

port19x:
Targeting a 4.15 release soon.
Testers welcome

port19x:
Known issues:
- iSH compatibility
- Botan installation on windows

port19x:
skip title doesn't only not work, it breaks ani-cli if you use that flag, no results found if you use it, but finds them if you don't

port19x (approved):
ready to merge imo, did extensive testing earlier.
Rest (windows, ios) can be done with minor releases / follow up PRs and/or in v5

port19x:
just for the record:
image

PR #1792, opened 2026-07-17
fix: update allanime_key, epoch and build_id

mbpowers:


DarioTD:
Thank you

port19x (approved):

port19x:
since you're just merging into the fix branch this is fine

happla:
This works for me also

DrStug:
ani-cli -U fix causes botan issues

Screenshot 2026-07-17 084914

port19x:
> ani-cli -U fix causes botan issues
> 
> Screenshot 2026-07-17 084914

because that's a dependency of the current fix.
It's not finalized

GreatestWaffle:
It appears this fix was working up until a about 1/2 hour ago, and it has now returned. Attempting to reinstall with the fix version still presents issue. Tested on Arch.

PR #1799, opened 2026-07-19
fix: implement aaReq AES-256-GCM without external botan dependency

Dhairya3391:
## Summary

- **Switches API endpoint** from `api.allanime.day` (behind Cloudflare) to `api.mkissa.net` (accessible via curl)
- **Adds dynamic key derivation** via a bootstrap endpoint instead of hardcoded keys
- **Keeps pure POSIX sh** — no external dependencies beyond what ani-cli already uses (curl, openssl, base64, od)
- **The existing `get_aa_req()` already implements AES-256-GCM** correctly via manual AES-CTR + GHASH — no changes needed there

## What changed

The upstream AllAnime API rotated its crypto scheme:
1. The API moved from `api.allanime.day` to `api.mkissa.net` (Cloudflare-free for the bootstrap/API endpoints)
2. The encryption key is now derived at runtime: `XOR(base64_decode(partB), mask)` where `partB` comes from `/client-crypto/v1/bootstrap?buildId=51` and `mask` is a constant embedded in the JS bundle
3. The bootstrap endpoint provides rotating `epoch` and `partB` values, keeping the key fresh

### Key derivation (new)
```
partB (base64 from bootstrap) → decode → XOR with 32-byte mask → AES-256 key
```
This produces the same `f34fa715...` key that was previously hardcoded, confirming the derivation is correct.

### What didn't need to change
- `get_aa_req()` — already implements AES-256-GCM manually (AES-CTR encryption + GHASH tag). It works with the new key as-is.
- `process_response()` — AES-CTR decryption works since the algorithm is the same, just the key changed.
- The persisted query hash `d405d0edd690624b66baba3068e0edc3ac90f1597d898a1ec8db4e5c43c00fec` still works on the new API.

### Build ID and mask rotation
The upstream periodically rotates the `buildId` and the XOR `mask` embedded in the JS bundle. The current values (as of 2026-07-20) are:
- `buildId`: **51**
- `mask`: `fe3bfee62898aaf5bbb0dbfed5f1cb33a59faab4ca5036885a42119a56429129`

When these rotate again, the script will need the updated values. The bootstrap endpoint (`/client-crypto/v1/bootstrap?buildId=N`) validates the build ID and returns an error for unknown ones, making it easy to detect when an update is needed.

## Testing

Verified end-to-end: search → episodes list → episode source URLs successfully fetched and decrypted via the mkissa API.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -249,106 +249,126 @@ process_response() {
     rm -f "$tmp"
 }
 
-hex_to_bin() { # $1=hex -> raw bytes on stdout (no xxd)
-    hex="$1"; len=${#hex}; i=0; oct=""
-    while [ $i -lt "$len" ]; do
-        pair=$(printf "%s" "$hex" | cut -c"$((i+1))-$((i+2))")
-        dec=$(printf "%d" "0x$pair")
-        oct="${oct}\\$(printf "%03o" "$dec")"
-        i=$((i+2))
+hex_to_bin() {
+    hb_h="$1"
+    hb_oct=""
+    while [ -n "$hb_h" ]; do
+        hb_t=${hb_h#??}
+        hb_p=${hb_h%"$hb_t"}
+        hb_d=$(( 0x$hb_p ))
+        hb_oct="$hb_oct\\$(( hb_d / 64 ))$(( (hb_d / 8) % 8 ))$(( hb_d % 8 ))"
+        hb_h="$hb_t"
     done
-    printf "%b" "$oct"
+    printf '%b' "$hb_oct"
 }
 
-xor_hex() { # $1 $2 hex strings (any length, $2 repeats if shorter) -> xor hex
-    awk -v a="$1" -v b="$2" '
-function hexval(c){
-    if (c>="0"&&c<="9") return c-0
-    c=tolower(c)
-    if (c=="a") return 10; if (c=="b") return 11
-    if (c=="c") return 12; if (c=="d") return 13
-    if (c=="e") return 14; if (c=="f") return 15
-    return 0
-}
-function xorbyte(x,y,   r,i,bx,by){
-    r=0
-    for(i=1;i<=128;i*=2){
-        bx=int(x/i)%2; by=int(y/i)%2
-        if (bx!=by) r+=i
-    }
-    return r
-}
-BEGIN{
-    la=length(a); lb=length(b); out=""
-    for(p=1;p<=la;p+=2){
-        h1=hexval(substr(a,p,1))*16+hexval(substr(a,p+1,1))
-        bp=((p-1)%lb)+1
-        h2=hexval(substr(b,bp,1))*16+hexval(substr(b,bp+1,1))
-        out=out sprintf("%02x", xorbyte(h1,h2))
-    }
-    printf "%s", out
-}'
+xor_hex() {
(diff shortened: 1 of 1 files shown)


71zenith:
@port19x What do we think about fetching the keys (epoch, build_id, etc) dynamically instead of hardcoding them. Considering they will eventually update them.
Ofcourse the chances of them switching to another encryption method is also there.

I have thought of two solutions. One is doing this at runtime incurring the cost of several(4-8 curl calls) to js chunks which can also very easily change syntax.
Other is to run a github workflow every few hours which we can easily control without needing to update the script.

Dojjino858:
Just a lurker, but I like this. Botan isn't a package available on windows from my understanding, meaning you will have to compile it and add it to your path.

port19x:
> @port19x What do we think about fetching the keys (epoch, build_id, etc) dynamically instead of hardcoding them. Considering they will eventually update them. Ofcourse the chances of them switching to another encryption method is also there.
> 
> I have thought of two solutions. One is doing this at runtime incurring the cost of several(4-8 curl calls) to js chunks which can also very easily change syntax. Other is to run a github workflow every few hours which we can easily control without needing to update the script.

I have no problem with updating via a github actions workflow, but am skeptical about doing it in-script, due to more network calls that way.
I'm waiting for the allanime situation to stabilize

ani-cli
```diff
@@ -249,6 +249,140 @@ process_response() {
...
+        pair=$(printf "%s" "$hex" | cut -c"$((i+1))-$((i+2))")
+        dec=$(printf "%d" "0x$pair")
+        oct="${oct}\\$(printf "%03o" "$dec")"
+        i=$((i+2))
+    done
+    printf "%b" "$oct"
+}
+
+xor_hex() { # $1 $2 hex strings (any length, $2 repeats if shorter) -> xor hex
+    awk -v a="$1" -v b="$2" '
```
port19x: cough cough

Dhairya3391: forgot it's in pr I was pushing to my branch

Dhairya3391: the perf improvement is good that is why I was pushing it to my main but pushed it here. my bad

Dhairya3391:
@port19x I think this should suffice if allanime don't change values again. still the values being changes frequently can be managed by just cron if they change we update it through workflow.

71zenith:
I don't think i was clear by the workflow approach. I meant as in running a workflow and storing the values in a gist or maybe in the repo itself and having the script fetch it. That way the user wouldn't have to manually update. This is indeed more complex than periodically updating the script. 
The PR itself looks good. I have tested on Windows and I would personally prefer this solution to botan even if its a bit of a black box to me.

port19x:
That would be sth like this: https://github.com/sdaqo/anipy-cli/tree/key-gen

port19x:
But tbh I'm skeptical.
Pivoting to a true crypto library like botan over openssl could pay off in the future.
We'd just have to move our current openssl stuff to botan too, so as to not have two dependencies for the same thing.

port19x:
A pivot to botan, fully replacing openssl, is very feasible.
See my latest comments in #1779

Dhairya3391:
i'll maintain this version without botan until ani-cli get's proper fix

PR #1801, opened 2026-07-20
fix: update allanime_key, epoch and build_id, set origin mkissa

mbpowers:


port19x (approved):
I'm not gonna quibble about the version number.
I know you changed it to appease our CI so that's fine.

PR #1828, opened 2026-07-27
feat: add anidb.app as fallback provider

armoox:
## Summary

Adds anidb.app as a fallback scraping provider. When allanime returns no sources, the script now falls back to anidb.app's public API for stream extraction.

### Changes
- Added `anidb_curl` with retry logic and `\r` stripping for Windows compatibility
- Added `search_anidb`, `episodes_list_anidb`, `get_anidb_m3u8` for anidb.app API
- Modified `get_episode_url` to try allanime first, then fall back to anidb.app if no links are found

### How it works
1. Allanime is tried first (existing behavior)
2. If allanime returns empty results, anidb.app is used as fallback
3. anidb.app uses its public API for search, episodes, and m3u8 stream extraction
4. Supports both sub and dub via language preference

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -308,22 +308,40 @@ get_aa_req() {
 
 # gets embed urls, collects direct links into provider files, selects one with desired quality into $episode
 get_episode_url() {
+    links=""
+
+    # try allanime first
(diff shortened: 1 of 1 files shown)


port19x:
Superseded by #1830

PR #1829, opened 2026-07-27
fix: new crypto flow

### LINKED ISSUE #1821: Illegal number: 0x
## **Make sure you are on the latest version by running `ani-cli -U`**

Version: 4.15.0
OS: Debian Linux 13 Trixie (ARM64)
Shell: fish
Anime: Chainsmoker Cat

**Describe the bug**
When attempting to download a show, ani-cli outputs the following message:
`/usr/bin/ani-cli: 253: Illegal number: 0x`

**Steps To Reproduce**

1. Run `ani-cli -d -e 1-4 "Chainsmoker Cat"`

**Expected behavior**
Ani-cli should download all specified episodes of the show.

Image

mbpowers:
did some deobfuscation !

changes on their end:
- added a new parameter called lane or k i.e. `"k": "k7"`
- need to request the part_b from their new crypto bootstrap endpoint requiring an `x-aa-boot` header

changes on our end:
- new regex for build_id and lane
- generate the mask and epoch locally, `generate_epoch()`, `generate_mask()` 
- `generate_aa_boot()` from mask, lane and epoch
- add `x-build-id` header
- add build_id and lane to aareq payload
- add lane to query extension

`generate_aa_boot()` uses open_ssl as my understanding is we would have used that had AES GCM been supported, but I believe open_ssl could be switched to botan if desired.

EDIT: openssl removed

I tested it for basic usecase but no further under the assumption things I didn't touch wouldn't have broken, in case anyone wants to test it more fully.

I think the regexes should be stable but we may want to wait for them to perform more builds to test against to see.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -269,10 +269,12 @@ EOF
 }
 
 generate_aa_boot() {
-    hmac_key=$(printf 'aa-boot:%s' "$allanime_build_id" | openssl dgst -sha256 -mac HMAC -macopt "hexkey:$allanime_mask" -binary)
-    hmac_key_hex=$(printf '%s' "$hmac_key" | od -An -tx1 -v | tr -d ' \n')
+    tmpdir="$(mktemp -d)"
+    trap 'rm -rf "$tmpdir"' EXIT INT
     payload="${allanime_build_id}:mkissa:mkissa.to:${allanime_epoch}${allanime_lane:+:$allanime_lane}"
-    printf '%s' "$payload" | openssl dgst -sha256 -mac HMAC -macopt "hexkey:$hmac_key_hex" | sed 's/^.*= //'
+    printf '%s' "$allanime_mask" | "$botan_exe" hex_dec - > "$tmpdir/key1.bin"
+    printf 'aa-boot:%s' "$allanime_build_id" | "$botan_exe" hmac --hash=SHA-256 "$tmpdir/key1.bin" | sed 's/ -//' | "$botan_exe" hex_dec - >"$tmpdir/key2.bin"
+    printf '%s' "$payload" | "$botan_exe" hmac --hash=SHA-256 "$tmpdir/key2.bin" | sed 's/ -//'
 }
 
 fetch_keys() {
--- a/ani-cli
+++ b/ani-cli
@@ -272,7 +272,7 @@ generate_aa_boot() {
     tmpdir="$(mktemp -d)"
     trap 'rm -rf "$tmpdir"' EXIT INT
     payload="${allanime_build_id}:mkissa:mkissa.to:${allanime_epoch}${allanime_lane:+:$allanime_lane}"
-    printf '%s' "$allanime_mask" | "$botan_exe" hex_dec - > "$tmpdir/key1.bin"
+    printf '%s' "$allanime_mask" | "$botan_exe" hex_dec - >"$tmpdir/key1.bin"
     printf 'aa-boot:%s' "$allanime_build_id" | "$botan_exe" hmac --hash=SHA-256 "$tmpdir/key1.bin" | sed 's/ -//' | "$botan_exe" hex_dec - >"$tmpdir/key2.bin"
     printf '%s' "$payload" | "$botan_exe" hmac --hash=SHA-256 "$tmpdir/key2.bin" | sed 's/ -//'
 }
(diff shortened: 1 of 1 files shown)


mbpowers:
I didn't bump the version since I wasn't sure if that's only for new features/api changes or what but lmk

andrempiva:
It's working for me. Thank you so much!

mbpowers:
okay so it seems like the values I hardcoded
```bash
    cat >"$tmp" <<'EOF'
10d13/213hI=
XdTA6FaAQSo=
KRCUT9lZh1o=
Aj3ASCIDMnw=
EOF
```
change each build. 

The issue is in the obfuscated js they look like:
```javascript
const nd = [
  Hr( - 518, - 584) + Hr( - 508, - 576),
  jr(854, 825) + jr(801, 821),
  jr(895, 833) + Hr( - 570, - 496),
  Hr( - 464, - 472) + jr(767, 818)
];
```
where those functions are
```javascript
function jr(e, t) {
  return Na(t - 614)
}
function Hr(e, t) {
  return Na(t - - 732)
}
function Na(e, t) {
  return e = e - 138,
  ml() [e]
}
function ml() {
  const e = [
    'table',
    'ype',
    'TUSln',
    // many many strings here...
    'rMopA',
    'ohONe',
    'deAt'
  ];
  return ml = function () {
    return e
  },
  ml()
}

(
  function (e, t) {
    const r = e();
    function n(i, s) {
      return Na(s - - 306)
    }
    function a(i, s) {
      return Na(s - 452)
    }
    for (; ; ) try {
      if (
        parseInt(n( - 164, - 144)) / 1 + parseInt(a(691, 615)) / 2 + parseInt(n( - 48, - 22)) / 3 + parseInt(n( - 92, - 166)) / 4 * (parseInt(n( - 120, - 62)) / 5) + parseInt(n( - 85, - 134)) / 6 * ( - parseInt(n( - 89, - 105)) / 7) + parseInt(a(654, 716)) / 8 + - parseInt(a(707, 664)) / 9 * (parseInt(n( - 117, - 162)) / 10) === t
      ) break;
      r.push(r.shift())
    } catch {
      r.push(r.shift())
    }
  }
) (ml, 101149);
```
so we would likely need to use node. I have a working [proof of concept](https://github.com/mbpowers/ani-extract) that is able to extract the needed functions from the AST using node and babel but that is more dependencies... If we want to remove babel I think it's possible to extract the relevant parts without parsing AST but would certainly be more fragile.

Thoughts?

The current values as of build 74 are
```bash
  ywI+GGWyMFA=
  ww8pcwjGfeY=
  8gjPB7mDWzc=
  nkCM5RxmdTY=
```
in case anyone needs them

hase0278:
Thanks for this working again. Hope they dont change the magic strings again soon.

Lost-Shadow99:
So is ani-cli  back in action?

mbpowers:
> So is ani-cli back in action?

this is currently working, but it still relies on hardcoded values that will eventually change. I have a way to scrape them but requires more than just sh, so it's unclear how best to proceed with a complete solution. I will update the hardcoded values if I see them change though, or you can scrape them yourself using [this](https://github.com/mbpowers/ani-extract)

Lost-Shadow99:
Sorry kinda dumb when it comes to this kind of code but how do I add this fix? And/or run this fix? I use the Linux terminal

mbpowers:
```sh
git clone https://github.com/mbpowers/ani-cli
cd ani-cli
git checkout crypto-fix
./ani-cli
```
note: this will not install it so you cannot call `ani-cli` anywhere, you must specify the path to this downloaded copy, as in `./ani-cli`, which is a relative path that works when inside the cloned repo dir.

Lost-Shadow99:
@mbpowers got it thx

Lost-Shadow99:
Hey so I'm trying to watch dragonball, but when I searched it all of the other versions accept the OG came up. What's that all about? Does the source being scraped simply not have it? The original version of ani-cli did.

mbpowers:
> Hey so I'm trying to watch dragonball, but when I searched it all of the other versions accept the OG came up. What's that all about? Does the source being scraped simply not have it? The original version of ani-cli did.

You need to search "dragon ball" not "dragonball". Nothing to do with these changes

port19x:
closing this since v5 is now released.
Thanks for providing a stopgap solution

PR #1837, opened 2026-07-29
fix: restore playback via AllAnime client-crypto bootstrap

### LINKED ISSUE #1821: Illegal number: 0x

ssnake:
## Problem

`ani-cli <anything>` fails at playback with:

```
ani-cli: 253: Illegal number: 0x
```

The source site (AllAnime / mkissa) replaced its **static** key material — a 64‑hex mask in the JS bundle plus `partB` in the landing HTML — with a runtime **client‑crypto bootstrap** handshake. Both scraped values are now empty, so the XOR loop in `fetch_keys` evaluated `0x` and aborted.

## Why the fix looks the way it does

The new key derivation only exists **after the site's obfuscated, self‑rotating JavaScript runs** (a runtime string‑array rotation defeats static grep/sed extraction, and the request‑signing proof `x-aa-boot` is an HMAC chain over a build‑id‑derived mask). It cannot be reproduced in POSIX shell.

`fetch_keys` now runs the site's **own** crypto in a JS runtime to obtain the AES‑256 key, epoch, build id and content lane:

- **deno** (preferred) — run with `--allow-net`, which confines the executed site code to network access only (no filesystem/process/env).
- **node ≥ 20** — fallback.

The embedded helper downloads the crypto chunk, neutralises its ESM imports, evaluates it in a stubbed browser‑like sandbox, and calls the site's own bootstrap entrypoint. It emits `KEY`/`EPOCH`/`BUILDID`/`LANE`, and `fetch_keys` `die`s with a clear message if anything fails (no more cryptic crash).

## Protocol changes the new backend requires

- `get_aa_req` payload is now `{v,ts,epoch,buildId,qh,k}`; IV = `SHA‑256(epoch:buildId:qh:ts:lane)[:12]`.
- `get_episode_url` request extensions include `"k":<lane>` and the request sends the `x-build-id` header.
- The persisted query hash is **unchanged**.

## New dependency (please review)

This adds a **deno or node** runtime requirement and **executes the source site's JavaScript**. That is a real departure from ani-cli's pure‑POSIX‑shell design, and I'm flagging it explicitly for discussion — but the site's key handshake is deliberately non‑reproducible in shell, so some form of JS execution is currently the only way to restore playback. `deno --allow-net` keeps the executed remote code sandboxed to the network.

## Verification

End‑to‑end against the live site: search → key derivation → source decryption → link resolution all succeed, and a real playable `.mp4` URL is returned (tested via `ANI_CLI_PLAYER=debug`). Helper produces identical output under both node and deno. `sh -n` and `bash -n` pass.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -227,33 +227,112 @@ select_quality() {
 }
 
 fetch_keys() {
-    page="$(mktemp)"
-    curl -s -A "$agent" "$allanime_refr" -o "$page" 2>/dev/null
+    # The source site derives its AES-256 key with an obfuscated, self-rotating
+    # "client-crypto" bootstrap handshake that cannot be reproduced in POSIX
+    # shell (the key material and the request-signing proof only exist after
+    # its JavaScript executes). We therefore run the site's *own* crypto in a
+    # JS runtime to obtain the key, epoch, build id and content lane. deno is
+    # preferred because it can be confined to the network with --allow-net;
+    # node is used as a fallback. See the aa_js_runtime dependency check.
+    keygen="$(mktemp)"
+    cat >"$keygen" <<'ANICLI_KEYGEN_JS'
+// ani-cli AllAnime client-crypto key deriver (runs under node >=20 or deno).
+// Args: <refr_url> <cdn_immutable_base>. Emits KEY/EPOCH/BUILDID/LANE lines.
+'use strict';
+var IS_DENO = typeof Deno !== 'undefined';
+var ARGV = IS_DENO ? Deno.args : process.argv.slice(2);
+var refr = ARGV[0], cdn = ARGV[1];
+var UA = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:150.0) Gecko/20100101 Firefox/150.0';
+var ENC = new TextEncoder();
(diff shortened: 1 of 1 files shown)


radumihai1:
works, tested with WSL Ubuntu 26.04
you have to install deno with:
curl -fsSL https://deno.land/install.sh | sh
then:
echo 'export DENO_INSTALL="$HOME/.deno"' >> ~/.bashrc
echo 'export PATH="$DENO_INSTALL/bin:$PATH"' >> ~/.bashrc
source ~/.bashrc

port19x:
I will leave this open as a potential fix, despite it duplicating the more native implementation in #1829

port19x:
closing this since v5 is now released.
Thanks for providing a stopgap solution

PR #1861, opened 2026-08-09
ci: claude pr review with fix suggestions and opt out

port19x:
# Pull Request Template

## Type of change

- [ ] Bug fix
- [x] Feature
- [x] Documentation update

## Description

Apparently this is possible.
I made sure to have it include instructions for committable suggestions in review comments, as that's the highest value automation we can get on top of our `shellcheck` and `shfmt` runs.
Contributors who aren't fond of AI are able to opt out of the review by ticking the box at the top of the PR template.

PR #1863, opened 2026-08-10
fix: correct user-level flatpak mpv path (io.mpv.Mpv, not io.mpv/Mpv)

TigerShark900:
## Summary
Fixes ani-cli failing to find/launch mpv on Fedora 44 Workstation (and any distro where mpv is installed via Flatpak rather than a native package). Fedora ships no `mpv` package by default — it needs RPM Fusion, which isn't enabled out of the box — so most Fedora users install mpv via Flatpak/GNOME Software instead. Two separate bugs in the Linux player-detection path stack on top of each other to break that case entirely:

**1. Wrong user-level flatpak path (`ani-cli` line ~410)**
The Linux fallback in b0993d4 added a correct system-wide flatpak path check (`/var/lib/flatpak/app/io.mpv.Mpv/`) but the existing user-level path (`$HOME/.local/share/flatpak/app/io.mpv/Mpv/`) has a typo: a slash where the real flatpak app-id uses a dot. Flatpak's real app-id directory is always `io.mpv.Mpv` (confirmed via `flatpak remote-info flathub io.mpv.Mpv` → `Ref: app/io.mpv.Mpv/x86_64/stable`, and the layout of other installed flatpaks on a real system, e.g. `app/com.brave.Browser/`). As written, `io.mpv/Mpv` can never exist, so a genuine user-level flatpak install of mpv is never detected.

**2. Missing case arm in the post-detection dependency check (`ani-cli` line ~513-519)**
Even once `player_function` correctly resolves to the flatpak path, a second, separate check (`dep_ch "$player_function"`) re-validates it with `command -v`, which fails on a literal directory path like `/var/lib/flatpak/app/io.mpv.Mpv/` — even though `play_episode()` already knows how to launch it via `flatpak run io.mpv.Mpv`. This check had no `*flatpak*mpv*` arm to skip past it, unlike `play_episode()`, so it always dies with `Program /var/lib/flatpak/app/io.mpv.Mpv/ not found. Please install it.`

Both bugs had to be fixed for flatpak-installed mpv to work at all on Linux, system-wide or per-user.

## Test plan
- [x] `sh -n ani-cli` passes
- [x] Reproduced and fixed live on Fedora 44 Workstation: installed mpv via `flatpak install flathub io.mpv.Mpv` (system-wide, matching GNOME Software's default), confirmed each error in sequence (`No player found...` → `Program /var/lib/flatpak/app/io.mpv.Mpv/ not found...`), fixed both, then ran `ani-cli` end-to-end and confirmed playback actually works
- [x] Simulated `dep_ch_failover` against both the correct (`io.mpv.Mpv`) and buggy (`io.mpv/Mpv`) user-level paths to confirm the exact failure mode before the fix
- [x] Verified the real flatpak app-id via `flatpak remote-info --user flathub io.mpv.Mpv`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -515,7 +515,6 @@ case "$player_function" in
     download) dep_ch_failover "yt-dlp,ffmpeg" >/dev/null || die 'Neither yt-dlp nor ffmpeg found' ;;
     android*) printf "\33[2K\rChecking of players on Android is disabled.\n" ;;
     *iSH*) printf "\33[2K\rChecking of players on iOS is disabled\n" ;;
-    *flatpak*mpv*) dep_ch "flatpak" ;;
     *) dep_ch "$player_function" ;;
 esac
 
--- a/ani-cli
+++ b/ani-cli
@@ -515,6 +515,7 @@ case "$player_function" in
     download) dep_ch_failover "yt-dlp,ffmpeg" >/dev/null || die 'Neither yt-dlp nor ffmpeg found' ;;
     android*) printf "\33[2K\rChecking of players on Android is disabled.\n" ;;
     *iSH*) printf "\33[2K\rChecking of players on iOS is disabled\n" ;;
+    *flatpak*mpv*) dep_ch "flatpak" ;;
     *) dep_ch "$player_function" ;;
 esac
 
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="5.0.1"
+version_number="5.0.2"
 
 # UI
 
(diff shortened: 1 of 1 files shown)


port19x:
Hey @TigerShark900 good catch!
Could you amend your commit to have Claude Code coauthor it?
We're still working on making an official policy around AI use, but coauthorship is definitely desired.

ani-cli
```diff
@@ -515,6 +515,7 @@ case "$player_function" in
     download) dep_ch_failover "yt-dlp,ffmpeg" >/dev/null || die 'Neither yt-dlp nor ffmpeg found' ;;
     android*) printf "\33[2K\rChecking of players on Android is disabled.\n" ;;
     *iSH*) printf "\33[2K\rChecking of players on iOS is disabled\n" ;;
+    *flatpak*mpv*) dep_ch "flatpak" ;;
```
port19x:
```suggestion
```

port19x (changes requested):
This one is not necessary, as flatpak availability is implicity checked via the dep_ch_failover function.

ani-cli
TigerShark900: Interesting, I've been getting an error without it: `Program /var/lib/flatpak/app/io.mpv.Mpv/ not found. Please install it.` even though mpv is installed. This is a separate issue from the incorrect directory fix. Without this line, `player_function` just falls through to the catch-all, which afterwards runs `command -v` on a raw directory path and always fails. If `dep_ch_failover` is working as intended, the problem is in `dep_ch()`'s catch-all not accounting for path-based results.

I can remove this line but then that catch-all becomes the problem again.

I set Claude Code as the coauthor as per your request.

port19x:
That first commit makes sense, but I don't see the purpose of explicitly checking for a flatpak binary when the very specific flatpak paths do exist

ani-cli
port19x: Yes, Claude Sonnet, the command -v fails and then the existence test for a folder or file succeeds right below it.
Read dep_ch_failover carefully

port19x:
Eyo wtf, shoutout to my own @Claude Opus for pointing out the line 518 default case that would otherwise run raw dep_ch on player function...
This is why I'm looking forward to #1861

ani-cli
port19x: Nvm 🫣

port19x (approved):

ani-cli
TigerShark900: yeahh

TigerShark900: Hey I just want to watch anime. I don't really care about a PR but I'm glad you figured it out. I don't code in shell or whatever
Sonnet signing out

port19x: That's valid, thx for spending your tokens with us XD

PR #1872, opened 2026-08-22
fix: forward skip_flag to syncplay player invocation

TheMridul:
## Summary
- `--skip` computes `skip_flag` via `ani-skip` and forwards it to the `*mpv*` and `*flatpak*mpv*` player branches in `play_episode()`, but the `*yncpla*` (syncplay) branch never included it, so combining `-s`/`--syncplay` with `--skip` silently did nothing.
- Adds `$skip_flag` to the syncplay invocation's player-args passthrough (after `--`), matching the other mpv-based branches.

## Test plan
- [x] Ran `ani-cli -s --skip "<anime>"` with `ani-skip` installed and confirmed intro is skipped when launched via syncplay.
- [x] Confirmed plain `ani-cli --skip "<anime>"` (no syncplay) still skips intros as before.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="5.0.2"
+version_number="5.0.3"
 
 # UI
 
--- a/ani-cli
+++ b/ani-cli
@@ -380,7 +380,7 @@ play_episode() {
             fi
             ;;
         *vlc*) nohup $player_function --http-referrer="$refr" $player_extra_flags --play-and-exit --meta-title="${anime_title} Episode ${ep_no}" "$video_link" ${sub_link:+:input-slave="$sub_link"} >/dev/null 2>&1 & ;;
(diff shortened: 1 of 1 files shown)


port19x:
I'm very interested in this change.
If you used AI in making this, which I don't want to accuse you of, but from the text it sounds a little like it, please add the used AI as a coauthor

TheMridul:
Bumped to 5.0.3 — thanks for catching that, I missed it in CONTRIBUTING.

On the AI question, fair to ask. The bug and the fix are mine: I hit it running `-s` together with `--skip`, noticed intro-skipping silently did nothing, and traced it to the syncplay branch being the only player branch that never forwarded `$skip_flag`. The one-line change is hand-written.

What I did put through an AI was the PR description, which is almost certainly what you picked up on. The version bump commit I just pushed was also AI-written, so that one carries a `Co-Authored-By` trailer. Happy to squash it into the fix or restructure the commits however you prefer.

port19x:
eyo wtf is copilot doing here??

port19x:
`./ani-cli -s -e 1 --skip "one piece"`

port19x:
closes: synacktraa/ani-skip#16

port19x:
Did a little mistake on the merge, but got it fixed quickly.
Thanks for the PR, will use this feature heavily 😄

port19x (approved):
Can confirm it works

Issue #1877, opened 2026-08-27
iOS iSH ‘No results found!’


Kvan2011:
## **Make sure you are on the latest version by running `ani-cli -U`**

**Metadata (please complete the following information)**
Version: 5.0.3
OS: iOS iSH
Shell: /bin/busybox
Anime: all?

**Describe the bug**
After the initial ‘ani-cli’ command, and asking to search, no matter which anime you search for it returns ‘No results found!’

**Steps To Reproduce**

1. Run `ani-cli`
2. Search any anime

**Expected behavior**
Should show a list of matching shows

**Screenshots**

<img width="1260" height="728" alt="Image" src="https://github.com/user-attachments/assets/3efc7889-7f92-41a0-9c0c-3dd92405bd39" />

**Additional context**

Lokendra2244:
same on arch linux
```
sudo ani-cli -U
[sudo] password for lokendra:
[branch:master] Checking for Update..
patching file /usr/bin/ani-cli
[branch:master] Updated: 5.0.2 -> 5.0.3
ani-cli --rofi
Checking dependencies...
xkbcommon: ERROR: [XKB-679] No Compose file for locale "en_IN.ISO8859-1": locale is either invalid or not installed
xkbcommon: ERROR: [XKB-679] couldn't find a Compose file for locale "en_IN" (mapped to "en_IN.ISO8859-1")
No results found!
```

ofbxyua-alt:
I'm getting the same thing on WSL

Checking dependencies...
Search anime: ten
No results found!

Ohanronnie:
same thing here!

TBA5854:
anidb.app is down

DavidsonRafaelK:
Some extra diagnostic data from macOS 15 (arm64, ani-cli 5.0.0 via Homebrew, also reproduced against the 5.0.3 code path).

anidb.app is not fully down — only the server-rendered HTML routes are. The JSON API is answering normally at the same moment:

```
$ curl_chrome136 -sS --max-time 20 -o /dev/null -w "%{http_code} %{time_total}\n" \
    "https://anidb.app/api/frontend/anime/777/episodes"
200 0.42

$ curl_chrome136 -sS --max-time 20 -o /dev/null -w "%{http_code} %{time_total}\n" \
    "https://anidb.app/browse?q=naruto"
000 20.00        # timed out, 0 bytes received
```

[...]

PR #1880, opened 2026-08-27
fix: report connection errors instead of "No results found!"

### LINKED ISSUE #1879: Network failures are reported as "No results found!", hiding connection errors
**Metadata**
Version: 5.0.3 (also reproduced on 5.0.0 from Homebrew)
OS: macOS 15 (arm64)
Shell: zsh / sh
Anime: any

**Describe the bug**

When `anidb_curl` fails at the network level — timeout, connection reset, empty body — `anidb_search` returns an empty string and the caller reports `No results found!`. That message tells the user their search term matched nothing, when in fact the site was never reached.

The Cloudflare guard in `anidb_search` only catches one specific failure shape:

```sh
_page="$(anidb_curl "$(printf "$search_api" "$1")" | tr '\n' ' ' | sed 's|<a href|\n<a href|g')"
if printf "%s" "$_page" | grep -qi "Just a moment"; then
    [ "$curl_exe" = "curl" ] && die "Blocked by cloudflare. Try installing curl-impersonate"
    die "Blocked by cloudflare."
fi
```

It matches the body string `Just a moment`. If curl exits non-zero and prints nothing — `--max-time 10` expiring is the common case — `$_page` is empty, the check does not fire, and execution reaches:

```sh
anime_list=$(anidb_search "$query")
[ -z "$anime_list" ] && die "No results found!"
```

So a site outage, a dropped connection, a DNS failure and a genuinely unknown anime all produce the same message.

**Steps to reproduce**

I hit this against the live anidb.app outage in #1877, where the HTML routes time out with 0 bytes while the JSON API still answers 200, and every search prints `No results found!`. Full route-by-route measurements are in https://github.com/pystardust/ani-cli/issues/1877#issuecomment-5438161564.

It reproduces deterministically by pointing the host at a blackhole address so the request times out the same way:

```sh
printf '198.51.100.1 anidb.app\n' | sudo tee -a /etc/hosts
ani-cli naruto
# -> "No results found!" after ~10s (--max-time 10 expiring), no hint that the host was unreachable
```

The underlying curl behaviour, without editing /etc/hosts:

```
$ curl -sL --max-time 10 --resolve anidb.app:443:198.51.100.1 "https://anidb.app/browse?q=x" -o /dev/null
$ echo $?
28          # timed out, empty stdout -> empty $_page -> "No results found!"
```

A refused connection takes the same path, just faster (`exit 7`, empty output). Remember to remove the /etc/hosts line afterwards.

**Expected behavior**

A network failure should say so. Something like `Connection error: could not reach https://anidb.app` lets the user tell "the site is down" apart from "no anime by that name", which is currently guesswork.

**Suggested fix**

The smallest change is an empty-response check right after the fetch, since an empty page is never a valid response here:

```sh
anidb_search() {
    #shellcheck disable=SC2059
    _page="$(anidb_curl "$(printf "$search_api" "$1")" | tr '\n' ' ' | sed 's|<a href|\n<a href|g')"
    [ -z "$_page" ] && die "Connection error: no response from $base_api"
    if printf "%s" "$_page" | grep -qi "Just a moment"; then
    ...
```

Note that propagating curl's exit status instead does not work as-is: `anidb_curl` runs inside a command substitution, so a `die` there exits only the subshell and the parent continues with an empty `$_page`. Checking the captured value in the caller avoids that.

`anidb_desc`, `anidb_episodes` and the m3u8 fetch use the same `anidb_curl` pattern and have the same blind spot, so the check is probably worth applying there too — `anidb_episodes` returning empty currently surfaces as an episode-selection failure rather than a network error.

This is filed separately from #1877 on purpose: that one is the anidb.app outage itself, which resolves on its own, while this is about the message ani-cli shows whenever any network failure happens. Happy to send a PR if you want it in this shape.

DavidsonRafaelK:
Closes #1879.

**Problem**

When `anidb_curl` fails at the network level the body is empty, so the `Just a moment` check in `anidb_search` never fires and the empty result falls through to `die "No results found!"`. A site outage, a dropped connection, a DNS failure and a genuinely unknown anime all produced the same message. That is what the current anidb.app outage in #1877 looks like from the user side.

There is a second half to it. Both scrapers are called inside a command substitution:

```sh
anime_list=$(anidb_search "$query")
```

so `die` only exits the subshell. The message is printed, the parent keeps going with an empty variable, and reaches the next `die`. A cloudflare block today prints both lines:

```
$ sh demo.sh          # die inside $( ), reduced to the essentials
Blocked by cloudflare. Try installing curl-impersonate
No results found!
```

**Change**

- Empty-response guard in `anidb_search` and `anidb_episodes`, the two scrapers where an empty body is never a legitimate answer. It names the host: `Connection error: no response from https://anidb.app`.
- `|| exit 1` at the four call sites, so the message the subshell printed actually stops the run instead of being followed by a second, contradictory one.
- Version bumped to 5.0.4.

`anidb_desc` is deliberately untouched: its metadata is optional and the caller already writes `anidb_desc "$anime_id" || true`. `process_hist_entry` is untouched too, since it runs backgrounded per history entry.

**Before / after**

Same binary, `base_api` pointed at a blackhole address (198.51.100.1) to force the timeout path:

```
=== 5.0.3 ===
Checking dependencies...
No results found!

=== patched ===
Checking dependencies...
Connection error: no response from https://198.51.100.1
```

**Checks**

- `shellcheck -s sh -o all -e 2250 ani-cli` — clean
- `shfmt -i 4 -ci -d ani-cli` — clean
- `test -x ani-cli` — ok
- no `awk`, no `wget` added
- `version_number` bumped, so the version-bump job sees it
- README needs no change: no flags, no behaviour outside the error path

Commit is co-authored by the agent that helped write it, as requested in #1879.

### DIFF
--- a/ani-cli
+++ b/ani-cli
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-version_number="5.0.3"
+version_number="5.0.4"
 
 # UI
 
@@ -157,6 +157,7 @@ anidb_curl() {
 anidb_search() {
     #shellcheck disable=SC2059
     _page="$(anidb_curl "$(printf "$search_api" "$1")" | tr '\n' ' ' | sed 's|<a href|\n<a href|g')"
+    [ -z "$_page" ] && die "Connection error: no response from $base_api"
     if printf "%s" "$_page" | grep -qi "Just a moment"; then
         [ "$curl_exe" = "curl" ] && die "Blocked by cloudflare. Try installing curl-impersonate"
         die "Blocked by cloudflare."
@@ -177,6 +178,7 @@ anidb_desc() {
 anidb_episodes() {
     #shellcheck disable=SC2059
     _page="$(anidb_curl "$(printf "$episodes_api" "${1##*-}")")"
+    [ -z "$_page" ] && die "Connection error: no response from $base_api"
     printf "%s" "$_page" | sed 's|},{|}\n{|g' | sed -nE 's|.*"id":([0-9]+).*"number":([0-9]+).*|\1\t\2|p'
 }
 
@@ -535,7 +537,7 @@ case "$source" in
         [ -z "$anime_id" ] && exit 1
         anime_title=$(printf "%s" "$anime_list" | grep "^$anime_id	" | cut -f 2 | sed 's| - episode.*||')
         anidb_desc "$anime_id" || true
-        episode_maps="$(anidb_episodes "$anime_id")"
+        episode_maps="$(anidb_episodes "$anime_id")" || exit 1
         ep_list="$(printf "%s" "$episode_maps" | cut -f 2)"
         ep_no=$(printf "%s" "$anime_list" | grep "^$anime_id	" | sed -nE 's/.*- episode (.+)$/\1/p')
         ;;
@@ -553,7 +555,7 @@ case "$source" in
 
         query=$(printf "%s" "$query" | sed 's| |+|g')
 
-        anime_list=$(anidb_search "$query")
+        anime_list=$(anidb_search "$query") || exit 1
         [ -z "$anime_list" ] && die "No results found!"
         [ "$index" -eq "$index" ] 2>/dev/null && result=$(printf "%s" "$anime_list" | sed -n "${index}p")
         [ -z "$index" ] && result=$(printf "%s" "$anime_list" | nl -w 2 | sed 's/^[[:space:]]//' | nth "Select anime: ")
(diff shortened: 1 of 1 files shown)


port19x:
Would be convenient rn, but doesn't seem to work
Screenshot 2026-09-05 at 13 50 45

port19x:
Dude, I just wanted to push a fix, why the fuck did you delete your repository

PR #1894, opened 2026-09-05
fix: temp-fix

Dhairya3391:
Fixes the HiAnime provider after recent server changes, cleans up subtitle handling, and resolves several edge-case bugs.

### Summary of Changes

- **Fix HiAnime Server Rename**: HiAnime renamed its video player server from HD-1 to ZokoAnime. The scraper now dynamically finds the ZokoAnime stream, with a fallback if server names change again.
- **Clean Search Results**: Strips out the site's 'Top 10' sidebar so searches only return matching titles instead of appending ~30 unrelated shows. Invalid queries now correctly show 'No results found!'.
- **Subtitles Without Temp Files**: Passes remote subtitle links directly to players (mpv, IINA, VLC, Syncplay) instead of downloading temp files. Colon-escapes URLs for IINA and uses `:input-slave` for VLC.
- **Subtitle Language Preference**: Selects English by default (configurable via `ANI_CLI_SUB_LANG`) instead of always grabbing the last track in the list.
- **Intro Skipping**: Ensures the anime ID is loaded before `ani-skip` runs so `--skip` works as expected.
- **History Fixes**: Validates the full title slug on HiAnime so older v5.0 history won't play the wrong show. Passes explicit arguments to background history workers to eliminate variable races.
- **Decimal Episode Numbers**: Uses exact text matching instead of regular expressions for episode numbers like `12.5` so the dot isn't treated as a wildcard.
- **Speed & Cleanup**: Replaces subprocess calls (`printf | sed`) with built-in shell parameter expansions, removes the unused description endpoint request, and increases curl max-time to 15s for large episode lists like One Piece.
- **Documentation**: Updates man pages and documentation files from anidb to hianime.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@
 </p>
 
 <h3 align="center">
-A cli to browse and watch anime (alone AND with friends). This tool scrapes the site <a href="https://anidb.app/">anidb.</a>
+A cli to browse and watch anime (alone AND with friends). This tool scrapes the site <a href="https://hianime.at/">hianime.</a>
 </h3>
 
 <h1 align="center">
@@ -534,12 +534,12 @@ Ani-skip uses the external lua script function of mpv and as such â€“ for now â€
 ## FAQ
 <details>
 	
-* Can I change subtitle language or turn them off? - No, the subtitles are baked into the video.
+* Can I change subtitle language or turn them off? - Subtitles are a separate english track handed to the player, so you can toggle them there (mpv: press `v`). Other languages are not offered.
 * Can I watch dub? - Yes, use `--dub`.
 * Can I change dub language? - No.
 * Can I change media source? - No (unless you can scrape that source yourself).
 * Can I use vlc? - Yes, use `--vlc` or `export ANI_CLI_PLAYER=vlc`.
-* Can I adjust resolution? - Yes, use `-q resolution`, for example `ani-cli -q 1080`.
+* Can I adjust resolution? - Yes, use `-q resolution`, for example `ani-cli -q 1080`. The provider currently serves a single 1080p stream, other values fall back to best.
 * How can I download? - Use `-d`, it will download into your working directory.
 * Can i change download folder? - Yes, set the `ANI_CLI_DOWNLOAD_DIR` to your desired location.
 * How can I bulk download? - `Use -d -e firstepisode-lastepisode`, for example `ani-cli onepiece -d -e 1-1000`.
--- a/ani-cli
+++ b/ani-cli
@@ -142,8 +142,10 @@ dep_ch_failover() {
 }
 
 cleanup() {
-    tput sgr0               # clears colors
-    rm -f "${histfile}.new" # remove temp logfile
+    tput sgr0                    # clears colors
+    rm -f "${histfile}.new"      # remove temp logfile
+    rm -f "${histfile}.subs"     # remove temp subtitle url
+    rm -f "${histfile}.vtt"      # remove temp subtitle file
 }
 
 # SCRAPING
@@ -180,7 +182,8 @@ deobfuscate_blob() (
         _char=$((_byte ^ _key))
         _output="${_output}\\$((_char / 64))$(((_char / 8) % 8))$((_char % 8))"
     done
-    printf "%b" "$_output" | sed -n 's|.*"src":"\([^"]*\.m3u8[^"]*\)".*|\1|p'
+    printf "%s\n" "$(printf "%b" "$_output" | sed -n 's|.*"src":"\([^"]*\.m3u8[^"]*\)".*|\1|p')"
+    printf "%s\n" "$(printf "%b" "$_output" | sed -n 's|.*"subtitles":\[[^]]*"src":"\([^"]*\.vtt[^"]*\)".*|\1|p')"
 )
 
 # search the query and give results. format is (id \t name)
@@ -219,7 +222,10 @@ hianime_m3u8() {
     _ep_id=$(printf "%s" "$episode_maps" | sed -n "s|^\([0-9][0-9]*\)\t${1}$|\1|p")
     _lang="sub"
     [ "$2" = "dub" ] && _lang="dub"
-    [ -z "$_ep_id" ] && return 1
+    [ -z "$_ep_id" ] && {
+        rm -f "${histfile}.subs"
+        return 1
+    }
 
     # 1. Get servers
     #shellcheck disable=SC2059
@@ -230,11 +236,17 @@ hianime_m3u8() {
     _hash="$(printf "%s\n" "$_servers" | grep "data-type=\"${_lang}\"" | grep "data-server-name=\"HD-1\"" | sed -n 's|.*data-hash="\([^"]*\)".*|\1|p' | head -n 1)"
     [ -z "$_hash" ] && _hash="$(printf "%s\n" "$_servers" | grep "data-type=\"${_lang}\"" | sed -n 's|.*data-hash="\([^"]*\)".*|\1|p' | head -n 1)"
 
-    [ -z "$_hash" ] && return 1
+    [ -z "$_hash" ] && {
+        rm -f "${histfile}.subs"
+        return 1
+    }
 
     # 3. Base64 decode embed URL
     _embed_url="$(b64_decode "$_hash")"
-    [ -z "$_embed_url" ] && return 1
+    [ -z "$_embed_url" ] && {
+        rm -f "${histfile}.subs"
+        return 1
+    }
 
     # Extract mal_id for ani-skip
     mal_id="$(printf "%s" "$_embed_url" | sed -n 's|.*/stream/mal/\([0-9][0-9]*\)/.*|\1|p')"
@@ -242,11 +254,24 @@ hianime_m3u8() {
     # 4. Fetch embed HTML
     _embed_page="$(hianime_curl -e "${base_api}/" "$_embed_url")"
     _blob="$(printf "%s" "$_embed_page" | sed -n 's|.*window\.__P="\([^"]*\)".*|\1|p')"
-    [ -z "$_blob" ] && return 1
+    [ -z "$_blob" ] && {
+        rm -f "${histfile}.subs"
+        return 1
+    }
 
     # 5. Deobfuscate
-    _m3u8_master="$(deobfuscate_blob "$_blob")"
-    [ -z "$_m3u8_master" ] && return 1
+    _deobfuscated="$(printf "%s\n" "$(deobfuscate_blob "$_blob")" | sed '/^$/d')"
+    _m3u8_master="$(printf "%s\n" "$_deobfuscated" | sed -n '1p')"
+    _subs="$(printf "%s\n" "$_deobfuscated" | sed -n '2p')"
+    [ -z "$_m3u8_master" ] && {
+        rm -f "${histfile}.subs"
+        return 1
+    }
+    if [ -n "$_subs" ]; then
+        printf "%s\n" "$_subs" >"${histfile}.subs"
+    else
+        rm -f "${histfile}.subs"
+    fi
 
     # 6. Fetch the master playlist and expand quality variants.
     _m3u8_content="$(hianime_curl -e "https://zokoanime.video/" "$_m3u8_master")"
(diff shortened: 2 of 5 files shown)


port19x:
Why is copilot AI here??? Wtf man

port19x:
@Dhairya3391 if this fixes the outage, bump it to v5.1 pls

port19x:
Oh, you brought your own copilot, that's fair

Dhairya3391:
@port19x anything else needed? it works as of now as fix but can't rely on this because site is not the best looking. 
edit : misclicked on tag

port19x:
@Dhairya3391 why ping pystardust lol.
We can keep this around for now, but we'll wait a little for if anidb comes back up after maintenance

port19x:
Please add your used AI tool as a coauthor, I'll give this a review pass in codex.
Too fat and too early for me to manually read all the code

71zenith:
I don't think anidb is coming back. This site also doesn't look very reliable. This is the other candidate that I could successfully scrape without being hls-blocked https://anizone.to/
But this is also another medium sized self hosted library and could go down as easily as anidb.

port19x:
Hey @71zenith, can you look at my PR over at #1895

port19x:
Want me to come into the discord again for the next provider?

Dhairya3391:
yeah we'll need to look at reliable providers again

71zenith:
I dont know how much productive it would be but we could communicate better i guess.

port19x:
I'm billionairelover

Tsagar3 (commented):
### Fix: subtitle tracks are dropped in the HiAnime flow

I tested this PR and found that **episodes play without subtitles**. Here's the root cause and a minimal fix.

#### Root cause

The HiAnime embed payload (the blurb decoded from `window.__P`) carries the subtitles **outside** the HLS playlist:

```json
{
  "download_url": "/download/mal/21/1/sub",
  "src": "https://hls2.aniwatchtv.uk/v/.../master.m3u8",
  "subtitles": [
    { "lang": "en", "label": "English", "default": true, "src": "https://hls2.aniwatchtv.uk/v/.../subs/6e4a0p3k8ehe9eee.vtt" }
  ],
  "sprite_vtt": ""
}
```

The `master.m3u8` (and the resulting `index.m3u8`) only contains `#EXT-X-STREAM-INF` variants — there are **no** `#EXT-X-MEDIA:TYPE=SUBTITLES` tracks. `ffprobe` on a segment confirms only `h264` video + `aac` audio.

`deobfuscate_blob` only extracts the stream `src` (the `.m3u8`), so the separate `.vtt` file referenced by `"subtitles"` is discarded. mpv then never receives a subtitle track, hence no subs.

#### Fix

* `deobfuscate_blob` also extracts the `.vtt` src from the `subtitles` array.
* `hianime_m3u8` persists it to a temp file (reusing the existing `histfile`-based tempfile pattern), cleared when the episode has no subs.
* `get_video_link` reads it into `sub_link`.
* `play_episode` passes `--sub-file="$sub_link"` to mpv (flatpak too) when present.

Verified with mpv against the extracted VTT:
`● Subs --sid=1 '6e4a0p3k8ehe9eee.vtt' (webvtt) [external]` and it fetches the `.vtt` with the correct `Referer`.

#### Shell quirk worth knowing

The decoded JSON has no trailing newline, and a lone `printf "%b" | sed -n 's/.../p'` may then emit its output **without** a trailing newline. Two consecutive extracts in a subshell can therefore concatenate into a single line (`master.m3u8https://...vtt`). The patch wraps each extract in `printf "%s\n" "$( ... )"` and filters blank lines downstream so the two URLs always land on their own lines.

#### Suggested diff

```diff
@@ cleanup() @@
 cleanup() {
-    tput sgr0               # clears colors
-    rm -f "${histfile}.new" # remove temp logfile
+    tput sgr0                    # clears colors
+    rm -f "${histfile}.new"      # remove temp logfile
+    rm -f "${histfile}.subs"     # remove temp subtitle url
 }

@@ deobfuscate_blob() @@
-    printf "%b" "$_output" | sed -n 's|.*"src":"\([^"]*\.m3u8[^"]*\)".*|\1|p'
+    printf "%s\n" "$(printf "%b" "$_output" | sed -n 's|.*"src":"\([^"]*\.m3u8[^"]*\)".*|\1|p')"
+    printf "%s\n" "$(printf "%b" "$_output" | sed -n 's|.*"subtitles":\[[^]]*"src":"\([^"]*\.vtt[^"]*\)".*|\1|p')"
 )

@@ hianime_m3u8() - "5. Deobfuscate" @@
-    _m3u8_master="$(deobfuscate_blob "$_blob")"
+    _deobfuscated="$(printf "%s\n" "$(deobfuscate_blob "$_blob")" | sed '/^$/d')"
+    _m3u8_master="$(printf "%s\n" "$_deobfuscated" | sed -n '1p')"
+    _subs="$(printf "%s\n" "$_deobfuscated" | sed -n '2p')"
     [ -z "$_m3u8_master" ] && return 1
+    if [ -n "$_subs" ]; then
+        printf "%s\n" "$_subs" >"${histfile}.subs"
+    else
+        rm -f "${histfile}.subs"
+    fi

@@ get_video_link() @@
     links=$(hianime_m3u8 "$ep_no" "$mode" | sort -g -r -s)
     [ -z "$links" ] && die "No sources found for $mode!"
     info "hianime.at links fetched"
+    sub_link="$(cat "${histfile}.subs" 2>/dev/null)"
     select_quality "$quality"

@@ play_episode() @@
     [ -z "$video_link" ] && get_video_link
+    sub_file_flag=""
+    [ -n "$sub_link" ] && sub_file_flag="--sub-file=$sub_link"
     # wait for the previous player to close (in range selection)
     wait
     # shellcheck disable=SC2086
     case "$player_function" in
-        debug) printf "All links:\n%s\nSelected link:\n%s\n" "$links" "$video_link" ;;
+        debug) printf "All links:\n%s\nSelected link:\n%s\nSubtitles:\n%s\n" "$links" "$video_link" "$sub_link" ;;
         ...
-        *flatpak*mpv*) flatpak run io.mpv.Mpv --http-header-fields="Referer: https://zokoanime.video/" $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link" >/dev/null 2>&1 & ;;
+        *flatpak*mpv*) flatpak run io.mpv.Mpv --http-header-fields="Referer: https://zokoanime.video/" $sub_file_flag $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link" >/dev/null 2>&1 & ;;
         *mpv*)
             if [ "$no_detach" = 0 ]; then
-                nohup $player_function --http-header-fields="Referer: https://zokoanime.video/" $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link" >/dev/null 2>&1 &
+                nohup $player_function --http-header-fields="Referer: https://zokoanime.video/" $sub_file_flag $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link" >/dev/null 2>&1 &
             else
-                $player_function --http-header-fields="Referer: https://zokoanime.video/" $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link"
+                $player_function --http-header-fields="Referer: https://zokoanime.video/" $sub_file_flag $skip_flag $player_extra_flags --force-media-title="${anime_title} Episode ${ep_no}" "$video_link"
```

Note: subs are currently only wired into the mpv paths; VLC and Syncplay would need `--sub-file` equivalents if so desired.

Happy to turn this into a commit on the branch if it helps.

Dhairya3391:
added sub fix, it's bit janky but works for now

jakobi:
Still a bug in deobfuscate, last line: 

The sed regex stretches is greedy, ignores languages of interest and thus extends to the end of array of subtitles: it always and only reports the vtt of the last-listed language:

```
printf "%s\n" "$(printf "%b" "$_output" | \
   sed -n 's|.*"subtitles" \[[^]]*"src":"\([^"]*\.vtt[^"]*\)".*|\1|p')"
```

To instead force english, this scrap works (plus some debug on stderr; line-splitting in sed really isn't my forte):

```
# match against label, do avoid capturing parens in the regex
export _langre="english|german" 
printf "%b" "$_output" | perl -e '
   undef $/; $_=<>; 
   s/.*"subtitles":\[([^]]*).*/$1/s and
     /"label":"[^"]*(?:$ENV{_langre}).*?"src":"([^"]+)/si and 
    printf "%s\n", "$1" and
    printf MAIN::STDERR "%s\n", "$1";
'
```
_Likely the language selection grep should be moved to the caller, rather than hidden in deobfuscate, i.e. deobfuscate in that case would return lines 2..n being the languages to egrep | head -1 against._

_Maybe an ordered list of regexes, (say ':' separated like in $PATH) would be better, as that would permit the user to specify the languages with an order of precedence._

OT: how does language selection work anyway in ani-cli, say for a dub? I didn't see anything when skimming for any use of $lang/$_lang or http headers (and $_lang is the dub/sub selection instead)?

Tsagar3:
Good catch @jakobi — I can confirm the greedy match. I reproduced it with a JSON mirroring the real embed (3 languages, English marked `"default":true`):

```
# current sed
.*"subtitles":\[[^]]*"src":"..."
        -> always returns the LAST-listed language (es.vtt in my test)
```

The problem is `[^]]*` swallows the whole array and the trailing `.*` backtracks to the last `"src":"` before the closing `]`, so language order (and `"default"`) is ignored. It just happened to work in my earlier test because One Piece ep 1 lists a single language.

The clean fix is to emit **one line per subtitle entry** and let the caller pick:

```diff
--- deobfuscate_blob
-    printf "%s\n" "$(printf "%b" "$_output" | sed -n 's|.*"subtitles":\[[^]]*"src":"\([^"]*\.vtt[^"]*\)".*|\1|p')"
+    printf "%s\n" "$(printf "%b" "$_output" | tr '}' '\n' | sed -n 's|.*"lang":"\([^"]*\)".*"label":"\([^"]*\)".*"src":"\([^"]*\.vtt[^"]*\)".*|\1\t\2\t\3|p')"
+
+# now emits: lang<TAB>label<TAB>src , one per language

--- hianime_m3u8 (deobfuscate step)
     _deobfuscated="$(printf "%s\n" "$(deobfuscate_blob "$_blob")" | sed '/^$/d')"
     _m3u8_master="$(printf "%s\n" "$_deobfuscated" | sed -n '1p')"
-    _subs="$(printf "%s\n" "$_deobfuscated" | sed -n '2p')"
+    _sub_entries="$(printf "%s\n" "$_deobfuscated" | sed -n '2,$p')"
+    _sub_langs="${ANI_CLI_SUB_LANG:-en}"
+    _subs="$(printf "%s\n" "$_sub_entries" | grep -iE "^$(printf '%s' "$_sub_langs" | tr ',:' '|')[[:space:]]" | head -n1 | cut -f 3)"
+    [ -z "$_subs" ] && _subs="$(printf "%s\n" "$_sub_entries" | head -n1 | cut -f 3)"
```

Selection logic (verified):
- picks the first entry whose `lang` matches `ANI_CLI_SUB_LANG` (default `en`, comma/colon-separated list gives ordered precedence, i.e. @jakobi's PATH-style idea for free)
- falls back to the first-listed entry when nothing matches (keeps the pre-fix default=true behaviour when English is listed first, like the real embed)
- everything downstream (`sub_link`, local VTT download, `--sub-file`/VLC/Syncplay flags) is untouched since it just consumes `_subs`

Re `grep -E`/case-insensitivity: the script already uses `grep -E` for quality selection, so this is consistent. If you want label-based matching too, extending the alternation to `label` is trivial.

And to answer your OT: language selection in ani-cli today is only sub vs dub — `--dub` sets `mode=dub`, which picks the dub server/stream via `data-type` on the servers list. The subtitle *language* itself wasn't user-selectable before this; `ANI_CLI_SUB_LANG` adds exactly that knob. For dub streams the `.vtt` is the dialogue subtitle track (usually `en`), which the default covers.

sader1992:
Why everyone want to overwrite anidb.app ? why not have it as an option ? like --provider (anidb/hianime) just incase , maybe Did anidb close forever ?

Also maybe make it like a plugin, so each provider would have a file for all it's codes and api.

PR #1895, opened 2026-09-05
feat: http error codes upon connection errors

port19x:
I now gotta open a separate PR because @DavidsonRafaelK had to close his #1880 while I worked on fixing his PR...


71zenith (approved):
Looks okay

PR #1902, opened 2026-09-11
refactor: deeper hianime_curl error reporting

port19x:
There was a mini outage where we didn't get those juicy HTTP codes for a second

Screenshot 2026-09-12 at 00 22 51
Screenshot 2026-09-12 at 00 23 53


port19x:
wait, I can compact this code a little

port19x:
nvm, that would make it less robust and way worse for readability

71zenith (approved):

PR #1911, opened 2026-09-13
fix: argument list too long on shells without a builtin printf

### LINKED ISSUE #1906: ani-cli in a 'Scriptrunner for Termux' script does not work with parameters since 5.1.1
Version: 5.1.1.
OS: Android 13
Shell: Termux
Anime: all

**Describe the bug**
Since 5.1.1 no longer able to call ani-cli in a bash script.

ani-cli works if typed into termux but if the same command is pulled from a .sh file, it results in the error:

/data/data/com.termux/files/usr/bin/ani-cli[601]: /data/data/com.termux/files/usr/bin/printf: Argument list too long

**Steps To Reproduce**

1. Run `ani-cli -d --dub flcl -r1--1`
2.  Create script.sh file
3. Add 'ani-cli -d --dub flcl -r1--1` and save
4.. bash script.sh
**Expected behavior**
Downloading will work only on step 1, not step 4

U-L-M-S:
## Type of change

- [x] Bug fix

## Description

Fixes #1906.

The reporter's shell is mksh, the `/bin/sh` of Android. It has no builtin `printf`, so `printf "%s" "$_response"` runs an external printf with the whole page as one argument. Linux allows 128KiB per argument, the search page is 250KiB, so it fails with `Argument list too long`.

Fix: a small `print_page` helper prints through a here-document. The shell writes that itself, so there is no limit. It is used only where whole pages are printed.

Why Termux users end up in mksh: `ani-cli -U` rewrites the package's shebang to `#!/bin/sh`, which is mksh on Android. That is a separate issue and not touched here.

Tested with mksh on Android (Termux) and in docker with mksh, dash, bash and busybox: master fails, this branch works.

## Checklist

- [x] any anime playing
- [x] bumped version
---
- [ ] next, prev, replay and select work
- [x] `-d` downloads work
- [x] `-S` select index works
- [x] `-q` quality works
- [ ] `-s` syncplay works (not installed)
- [ ] `-v` vlc works (not installed)
---
- [x] Readme is up to date
- [x] Man page is up to date


port19x:
Closing this because it is caused by user error in my opinion as laid out in the linked issue.

PR #1914, opened 2026-09-16
feat: add exact title matching flag

### LINKED ISSUE #1851: feat: --exact flag to only match if exact full string is found (for wrapper scripts)
This is related to my problem for #1850. I used a wrapper script which calls ani-cli to play a specific episode from a specific show, but since only one result appeared for the show I searched and it was the wrong show, it was auto selected and started playing without any warning that it wasn't an exact match. 

 For scripting it would be nice to have some flag to force an exact match against the titles of the search results (`--exact` or `--exact-match`).

Voyagerroc-Lab:
## Summary
- add --exact to require an exact, case-insensitive anime title match
- prevent wrapper scripts from silently selecting an unrelated search result
- bump the script version and document the new option in help output

Closes #1851

## Validation
- git diff --check
- POSIX shell checks were unavailable because sh, shellcheck, and shfmt are not installed on this Windows environment.


port19x:
Can we extend this to include an environment variable `ANI_CLI_EXACT` ?

ani-cli
```diff
@@ -488,6 +490,7 @@ histfile="$hist_dir/ani-hsts"
 
 source="${ANI_CLI_DEFAULT_SOURCE:-search}"
 branch="${ANI_CLI_BRANCH:-master}"
+exact_match=0
```
U-L-M-S, line 493: This is the `ANI_CLI_EXACT` env var port19x asked for. `--exact` still wins because the argument loop below runs after this line, same as `ANI_CLI_SKIP_INTRO` and `--skip`.

```suggestion
exact_match="${ANI_CLI_EXACT:-0}"
```

ani-cli
```diff
@@ -607,9 +611,13 @@ case "$source" in
...
         [ "$source" = "nextep" ] && time_until_next_ep "$query"
 
+        search_query="$query"
         query=$(printf "%s" "$query" | sed 's| |+|g')
 
         anime_list=$(hianime_search "$query") || exit 1
+        if [ "$exact_match" -eq 1 ]; then
+            anime_list=$(printf "%s\n" "$anime_list" | awk -F '\t' -v title="$search_query" 'tolower($2) == tolower(title)')
+        fi
         [ -z "$anime_list" ] && die "No results found!"
```
U-L-M-S, line 621: Two things here:
1. CI fails on **No Awk Allowed**: `.github/workflows/ani-cli.yml` greps `ani-cli` for `awk` and rejects it.
2. The argument loop has already replaced spaces with `+`, so `search_query` is `one+piece` and can never equal `One Piece`. With the current commit, `ani-cli --exact one piece` prints "No results found!" even though One Piece is in the results.

This version stays POSIX (`read`, `cut`, `tr`, `grep -ixF`, so titles containing `[`, `.`, `*` or `;` are matched literally), turns `+` back into spaces on both sides before comparing, and gives a distinct error when the search had results but none matched exactly. It passes `shellcheck -s sh -o all -e 2250` and `shfmt -i 4 -ci -d`.

```suggestion
        query=$(printf "%s" "$query" | sed 's| |+|g')

        anime_list=$(hianime_search "$query") || exit 1
        [ -z "$anime_list" ] && die "No results found!"
        if [ "$exact_match" = 1 ]; then
            # spaces in the query were replaced with +, undo that on both sides so titles compare as typed
            exact_title=$(printf "%s" "$query" | tr '+' ' ')
            anime_list=$(printf "%s\n" "$anime_list" | while IFS= read -r line; do
                printf "%s" "$line" | cut -f 2 | tr '+' ' ' | grep -qixF -- "$exact_title" && printf "%s\n" "$line"
            done)
            [ -z "$anime_list" ] && die "No exact match for \"$exact_title\"!"
        fi
```

U-L-M-S (commented):
Tested locally with the two suggestions applied (stubbed search plus live runs with `ANI_CLI_PLAYER=debug`): `--exact one piece` plays One Piece, `--exact "one piec"` and `ANI_CLI_EXACT=1 ani-cli one piec` refuse with "No exact match", and behaviour without the flag is unchanged. All repo checks pass (shellcheck, shfmt, no awk, version bump).

`ani-cli.1` is not part of this PR, so no inline suggestion for it. The matching man page entries would be, under OPTIONS (after `--select-nth`):

```
.TP
\fB\--exact\fR
Only accept search results whose title exactly matches the query (case-insensitive). Useful for wrapper scripts, so a single unrelated result is not selected automatically.
```

and under ENVIRONMENT VARIABLES:

```
.TP
\fBANI_CLI_EXACT\fR
Controls if only search results whose title exactly matches the query (case-insensitive) are accepted, equivalent to --exact. Can be 0 (disabled) or 1 (enabled). Default is 0.
```

